PR #1, opened 2018-10-09
Don't allow puting python objects into a directory style path object

stevededalus:


ocean/setup.txt
```diff
@@ -13,7 +13,6 @@ chardet==3.0.4            # via requests
 docutils==0.14            # via botocore
 elasticsearch==6.3.1
 future==0.16.0
-futures==3.2.0            # via pyarrow, s3transfer
```
dimaryaz: Is this intentional? (I'm not actually sure how setup.txt is used)

stevededalus: No, thanks for catching. Reverted that file.

PR #2, opened 2018-10-09
Add gotchas and metadata search to user documentation

ResidentMario:


UserDocs.md
```diff
@@ -256,7 +253,11 @@ To modify which file types are searchable, populate a `.quilt/config.json` file
...
 ```
 
+Full-text search will include metadata matches in the results. To search metadata only, enter a string of the form `user_meta.field:"value"` into the search field, replacing `field` with the name of the field of interest and `value` with its (quoted) value.
+
 ## Known limitations and gotchas
 
-* At present, only objects placed into an S3 bucket via the T4 API are searchable. Existing objects in the S3 bucket, and objects placed in the bucket via the S3 API, are not.
-* At present, due to limitations with ElasticSearch indexing, we do not recommend including indexing files that are over 10 MB in size.
\ No newline at end of file
+* Only objects placed into an S3 bucket via the T4 API are searchable. Existing objects in the S3 bucket, and objects placed in the bucket via the S3 API, are not.
```
akarve: i rewrote this bullet for correctness in #5

PR #10, opened 2018-10-11
Add warning about trailing forward slash

meffij:


ResidentMario:
Two comments:

1. This probably belongs in the Issues section, alongside the rest of the warnings about rough edges.
2. As I understand it the problem isn't pushing anything ending in `/` to T4. The problem is pushing *files* ending in `/` to T4. If I push a *directory* to a key ending in `/`, then everything works fine. And indeed we do this, and recommending doing this, in `QuickStart.ipynb`.

@akarve what are your thoughts?

UserDocs.md
```diff
@@ -38,6 +38,8 @@ Writes the in-memory object `src` to the path `dest` in S3.
 Transparently serializes some object types.
 See [Serialization](#serialization) for details.
 
+`dest` should not end in `/`. Objects that end in `/` are treated specially by some s3 tools in a way that is potentially dangerous, so it's best to avoid them.
```
akarve: not the right place to document this - because `he.put()` automatically forbids it

UserDocs.md
```diff
@@ -55,6 +57,8 @@ Writes the in-memory object `src` to the path `dest` in S3.
...
```
akarve: Also not the right place to document this as the API only allows `dest/` if you have `src/` and `src/` is a dir? if that's not the case we should make it the case :)

UserDocs.md
```diff
@@ -55,6 +57,8 @@ Writes the in-memory object `src` to the path `dest` in S3.
...
+
```
akarve: My general feedback. This bullet belongs in the "Known Issues" section, with a link to the corresponding AWS docs. Does not belong in API docs b/c API should forbid this case whenever possible?

akarve:
@ResidentMario agreed - just check that `put_file(src, dest/)` throws an exception

UserDocs.md
```diff
@@ -353,3 +353,6 @@ T4 automatically populates the following metadata:
     s3:GetObject
     s3:GetObjectVersion
     ```
+* The keys of objects in s3 should not end in `/`. Objects that end in `/` are treated specially by some s3 tools in a way that is potentially dangerous, so it's best to avoid them. The helium API will help you avoid this rough edge by rejecting keys that end in `/`.
```
ResidentMario: We capitalize `s3` as `S3` throughout the rest of the doc, we should do that here too.

UserDocs.md
```diff
@@ -353,3 +353,6 @@ T4 automatically populates the following metadata:
...
+
+Refer to [Amazon's documentation](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/using-folders.html) on folder objects.
```
ResidentMario: I would make this the last sentence in the paragraph above. Right now it's hanging out on its own line below the rest of the text in the rendered text on GitHub.

UserDocs.md
```diff
@@ -353,3 +353,8 @@ T4 automatically populates the following metadata:
     s3:GetObject
     s3:GetObjectVersion
     ```
+* The keys of objects in S3 should not end in `/`. Objects that end in `/`
```
akarve: "Objects whose keys end in `/`

UserDocs.md
```diff
@@ -353,3 +353,8 @@ T4 automatically populates the following metadata:
...
+are treated specially by some S3 tools in a way that
+is potentially dangerous, so it's best to avoid them.
+The helium API will help you avoid this rough edge by rejecting keys that end in `/`.
```
akarve: "by rejecting object keys that end in `/`"
(i.e. folder keys that end in `/` are legit and we should probably spell that out)

PR #14, opened 2018-10-11
config() updated with nicer repr, docs updated

eode:
This is a very lightweight class used for interpreter display when a user calls `he.config()`.  This is similar in spirit to our `DisplayList`, but doesn't require usage of HTML.

Quite a few options were considered, but a lightweight class made the most sense -- we want to retain order, and we want to display the filename.

* returning a tuple is obscure and breaks consistency
* returning a namedtuple breaks consistency, and the contained config doesn't display nicely
* returning an OrderedDict is just ugly, and doesn't include the filename
* returning a regular dict displays fine, but loses order and doesn't include the filename

### DIFF
--- a/UserDocs.md
+++ b/UserDocs.md
@@ -91,13 +91,15 @@ If `dest` contains file that `src` does not, that fill will appear as an Add.
 ### Configuration
 
 #### `helium.config()`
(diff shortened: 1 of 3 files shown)


eode:
Is a simple table-based `_repr_html_` desirable for jupyter display?  ..something like this?

screenshot from 2018-10-12 12-48-43

PR #25, opened 2018-10-18
Snapshots prototype

meffij:
Strawman API to be implemented by the 26th

TODO:
- [x] add todos
- [x] ~"snapshot"~ => "package"


ocean/helium/snapshots.py
```diff
@@ -138,3 +138,69 @@ def download_bytes_from_snapshot(src, snapshothash):
...
+            value: string
+        }
+        size is the length of the object in bytes
+        """
+        self._data = {}
+        pass
+
+    @staticmethod
+    def load(path):
+        """ loads a snapshot from a path """
```
stevededalus: Let's make all the docstrings sentences, capital letters to start and punctuation to end.

ocean/helium/snapshots.py
```diff
@@ -138,3 +138,69 @@ def download_bytes_from_snapshot(src, snapshothash):
...
+
+    def update(self, local_key, entry):
+        """ returns a new snapshot with the object at local_key set to entry """
+        raise NotImplementedError
+
+    def delete(self, local_key):
+        """ returns a new snapshot with local_key removed """
+        raise NotImplementedError
+
+    def tophash(self):
```
stevededalus: tophash is two words s/tophash/top_hash

ocean/helium/snapshots.py
```diff
@@ -138,3 +138,69 @@ def download_bytes_from_snapshot(src, snapshothash):
...
+
+    def get_file(self, logical_key, local_path):
+        """ gets object from local_key inside the snapshot and saves it to local_path """
+        raise NotImplementedError
+
+    def dump(self, path):
+        """ serializes this snapshot to a file at path """
+        raise NotImplementedError
+
+    def update(self, local_key, entry):
```
stevededalus: I think this should take a mapping directly, either another dictionary object or an iterable of key:value pairs
`mysnapshot.update({'fookey':'foovalue', 'barkey':'barvalue'})`

ocean/helium/snapshots.py
```diff
@@ -138,3 +138,69 @@ def download_bytes_from_snapshot(src, snapshothash):
...
+        """
+        returns the tophash of the snapshot.
+        Note that physical keys are not hashed because the snapshot has
+            the same semantics regardless of where the bytes come from.
+        """
+        raise NotImplementedError
+
+    def materialize(self, path):
+        """
+        copies each object in this snapshot to path according to logical key structure,
```
stevededalus: Let's adopt the google style/PEP257 standards.

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings

So, one line description, blank line, in depth description., blankline, "args" section, "returns" section and "raises" section

ocean/helium/snapshots.py
```diff
@@ -138,3 +138,69 @@ def download_bytes_from_snapshot(src, snapshothash):
...
+
+    def tophash(self):
+        """
+        returns the tophash of the snapshot.
+        Note that physical keys are not hashed because the snapshot has
+            the same semantics regardless of where the bytes come from.
+        """
+        raise NotImplementedError
+
+    def materialize(self, path):
```
stevededalus: is this necessary when we can call 'get_files(*)' or 'get_files('/')'. I also prefer 'fetch' or 'install' over materialize.

meffij: The difference here is that `materialize` generates a new snapshot with physical keys pointing to the new objects

ocean/helium/snapshots.py
```diff
@@ -138,3 +143,263 @@ def download_bytes_from_snapshot(src, snapshothash):
...
+            A new snapshot
+
+        Raises:
+            KeyError: when logical_key is not present to be deleted
+        """
+        snap = self._clone()
+        snap._data.pop(logical_key)
+        return snap
+
+    def top_hash(self):
```
stevededalus: should this be a private function? is it computing the snapshot hash or just returning it from the metadata?

ocean/helium/snapshots.py
```diff
@@ -138,3 +143,263 @@ def download_bytes_from_snapshot(src, snapshothash):
...
+        Args:
+            logical_key: logical key to update
+            entry: new entry to place at logical_key in the snapshot
+
+        Returns:
+            A new snapshot
+        """
+        snap = self._clone()
+        # TODO validate entry contents
+        snap._data[logical_key] = entry
```
stevededalus: If the value is None then remove it

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,267 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
     snapshot_data = read_snapshot_by_hash(bucket, snapshothash)
     obj_rec = snapshot_data['contents'][key]
     return download_bytes(src, version=obj_rec['VersionId'])
+
+def hash_file(path):
+    """ Returns SHA256 hash of file at path. """
+    with open(path, 'rb') as f:
+        return hashlib.sha256(f.read()).hexdigest()
+
+class PackageException(Exception):
```
stevededalus: Let's move this into an `exceptions.py` module before landing.

ocean/helium/snapshots.py
meffij: It will return the top hash, which will be calculated lazily by a private function

ocean/helium/snapshots.py
meffij: that's delete -- if you want we can merge that into one API call, but right now delete is the clear way to remove logical_keys from a package

ocean/helium/snapshots.py
meffij: OK

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,267 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
 def download_bytes_from_snapshot(src, snapshothash):
     bucket, key = split_path(src)
     snapshot_data = read_snapshot_by_hash(bucket, snapshothash)
     obj_rec = snapshot_data['contents'][key]
     return download_bytes(src, version=obj_rec['VersionId'])
+
+def hash_file(path):
+    """ Returns SHA256 hash of file at path. """
+    with open(path, 'rb') as f:
+        return hashlib.sha256(f.read()).hexdigest()
```
dimaryaz: Do it in chunks - don't want to load the whole file into memory

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,267 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        Raises:
+            file not found
+            json decode error
+            invalid package exception
+        """
+        data = {}
+        with jsonlines.open(path) as reader:
+            meta = reader.read()
+            for obj in reader:
+                lk = obj['logical_key']
```
dimaryaz: `lk = obj.pop('logical_key')`

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,267 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        files = src_path.rglob('*')
+        for f in files:
+            if not f.is_file():
+                continue
+            entry = {
+                'hash': hash_file(f),
+                'size': os.path.getsize(f),
+                'user_meta': {},
+                'physical_keys': [{
+                    'type': 'local_file',
```
dimaryaz: Should probably use an `Enum`

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,267 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+                    'path': os.path.abspath(f)
```
dimaryaz: Should probably be relative to `path`, or some other root?

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,267 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+            KeyError: when logical_key is not present in the package
+            physical key failure
+            hash verification fail
+        """
+        entry = self._data[logical_key]
+        physical_keys = entry['physical_keys']
+        physical_key = physical_keys[0] # TODO: support multiple physical keys
+        stream = dereference_physical_key(physical_key)
+        # TODO: verify hash
+        if 'target' in entry:
```
dimaryaz: `stream` is now an open file descriptor... and it'll get leaked if anything happens after this.

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,267 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+
+        Raises:
+            fail to create file
+            fail to finish write
+        """
+        with open(path, mode='w') as f:
+            with jsonlines.Writer(f) as writer:
+                writer.write(self._meta)
+                for logical_key, obj in self._data.items():
+                    obj['logical_key'] = logical_key
```
dimaryaz: You modifying `self._data` here... probably don't want that

ocean/helium/snapshots.py
meffij: Why wouldn't it be absolute? This is for the physical key

ocean/helium/snapshots.py
meffij: correct

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,282 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+            for obj in reader:
+                lk = obj.pop('logical_key')
+                if lk in data:
+                    raise PackageException("Duplicate logical key while loading package")
+                data[lk] = obj
+
+        return Package(data, meta)
+
+    @staticmethod
+    def pkg(path):
```
stevededalus: maybe `build_package`

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,282 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        raise NotImplementedError
+
+    def get_meta(self, logical_key):
+        """
+        Returns user metadata for specified logical key.
+        """
+        entry = self._data[logical_key]
+        return entry['user_meta']
+
+    def dump(self, path):
```
stevededalus: s/dump/save_manifest or s/dump/build_manifest or something more descriptive

ocean/helium/snapshots.py
```diff
@@ -131,10 +136,282 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        physical_key = physical_keys[0] # TODO: support multiple physical keys
+        stream = dereference_physical_key(physical_key)
+        # TODO: verify hash
+        if 'target' in entry:
+            # TODO: dispatch on target to deserialize
+            raise NotImplementedError
+
+        return stream
+
+    def get_file(self, logical_key, path):
```
stevededalus: s/get_file/get_files

ocean/helium/snapshots.py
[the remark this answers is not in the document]
meffij: It creates a logical key that corresponds to the place of the file in the directory structure. For example, if you `create_package('.')` in a folder with one file, `a.json`, then create_package will create one logical key, `a.json`. If you go a directory up, the logical key will be `some_dir/a.json`. It uses the filesystem to infer logical keys essentially.

api/python/t4/snapshots.py
```diff
@@ -131,10 +138,283 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+                writer.write(self._meta)
+                for logical_key, obj in self._data.items():
+                    writer.write({**{'logical_key': logical_key}, **obj})
+
+    def update(self, logical_key, entry):
+        """
+        Returns a new package with the object at logical_key set to entry.
+
+        Args:
+            logical_key: logical key to update
```
stevededalus: Throw the types in the Args doc string `logical_key(string)` I assume. I guess entry is a dict, but we'll probably want to make it it's own class.

api/python/t4/exceptions.py
```diff
@@ -0,0 +1,6 @@
+""" Exceptions for Helium """
```
dimaryaz: T4 :)

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+
+        Raises:
+            KeyError: when logical_key is not present in the package
+            physical key failure
+            hash verification fail
+            when deserialization metadata is not present
+        """
+        entry = self._data[logical_key]
+        physical_keys = entry.physical_keys
+        physical_key = physical_keys[0] # TODO: support multiple physical keys
```
dimaryaz: Maybe `raise NotImplementedError`, just in case?

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+    raise NotImplementedError
+
+def get_package_registry(path):
+    """ Returns the package registry for a given path """
+    if path.startswith('s3://'):
+        bucket = path[5:].partition('/')[0]
+        return "s3://{}/.quilt/packages/".format(bucket)
+    else:
+        # local path
+        return "appdirs/.quilt/packages" # TODO: replace with actual logic
```
dimaryaz: `raise NotImplementedError`; safer than saving stuff in an unexpected location

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+
+        Raises:
+            fail to create file
+            fail to finish write
+        """
+        with open(path, mode='w') as f:
+            with jsonlines.Writer(f) as writer:
+                writer.write(self._meta)
+                for logical_key, entry in self._data.items():
+                    writer.write({**{'logical_key': logical_key}, **entry.as_dict()})
```
dimaryaz: I think you can just do:
`writer.write({'logical_key': logical_key, **entry.as_dict()})`

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+
+        Returns:
+            None
+
+        Raises:
+            fail to create file
+            fail to finish write
+        """
+        with open(path, mode='w') as f:
+            with jsonlines.Writer(f) as writer:
```
dimaryaz: Maybe it's too early to worry about this just yet, but JSON won't scale well to 1,000,000 files. We should think about using binary formats.

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        Args:
+            logical_key(string): logical key to update
+            entry(PackageEntry): new entry to place at logical_key in the package
+
+        Returns:
+            A new package
+        """
+        pkg = self._clone()
+        # TODO validate entry contents
+        pkg._data[logical_key] = entry
```
dimaryaz: I see so many places that add/remove `logical_key` to/from entry. Would it make sense to add another layer of nesting, and not have `logical_key` be a part of that dict?

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        Returns:
+            None
+        """
+        raise NotImplementedError
+
+    def top_hash(self):
+        """
+        Returns the top hash of the package.
+
+        Note that physical keys are not hashed because the package has
```
dimaryaz: This seems a bit strange to me. It sometimes hashes the package, sometimes doesn't?

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        return self._meta['top_hash']
+
+    def textual_hash(self):
+        """
+        Returns the textual hash of the package manifest serialized to disk
+
+        Returns:
+            SHA256 hash of serialized manifest
+        """
+        with tempfile.TemporaryDirectory() as tmpdir:
```
dimaryaz: Why not a `NamedTemporaryFile`?

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        Raises:
+            fail to get bytes
+            fail to put bytes
+            fail to put package to registry
+        """
+        if name is None:
+            name = self.textual_hash()
+        self.get_files(path)
+        self.dump(get_package_registry(path) + name)
+        raise NotImplementedError
```
dimaryaz: Just fail immediately; it's weird to do some work, then fail

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+            size(number): size of object in bytes
+            hash({'type': string, 'value': string}): hash object
+            meta(dict): metadata dictionary
+
+        Returns:
+            a PackageEntry
+        """
+        self.physical_keys = physical_keys
+        self.size = size
+        self.hash = hash_obj
```
dimaryaz: `self.hash_obj = hash_obj`

I'd expect the current code to fail somewhere? Is it not being used?

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        }
+        return copy.deepcopy(ret)
+
+class Package(object):
+    """ In-memory representation of a package """
+
+    def __init__(self, data=None, meta=None):
+        """
+        _data is of the form {logical_key: PackageEntry}
+        physical_keys is a list of objects of the form {
```
dimaryaz: Move this to `PackageEntry`?

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+
+    def textual_hash(self):
+        """
+        Returns the textual hash of the package manifest serialized to disk
+
+        Returns:
+            SHA256 hash of serialized manifest
+        """
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tmp_name = tmpdir + 'temp.pkg'
```
dimaryaz: Actually, this creates a world-readable file outside of the temp dir cause you don't have a `/`

api/python/t4/snapshots.py
eode: I don't know -- I actually think this is *really* the time to be thinking about this, before we've built a lot of inertia in a particular direction.

This is an example of one of the reasons I'd like to be using pluggable interfaces for a lot of our accessing of external resources -- like:

* `PathHandlerABC`, which gets implemented into `PathHandlerFS`, `PathHandlerS3`, `PathHandlerHTTP`, etc for browsing arbitrary file trees.
* `ObjectHandlerABC`, which gets implemented into `ObjectHandlerFS`, `ObjectHandlerS3`, `ObjectHandlerDAVFS`, or whatever, for putting/getting binary blobs+metadata,
* `PackageHandlerABC` which gets implemented into `PackageHandlerS3`, `PackageHandlerFS`
* `SerDesABC` -- you get the idea.

..in any case, this is actually using JSON Lines, which can be done incrementally (although I don't know if it has been, here) so should scale fine for dumping.  But, I completely agree with the underlying point.

If we have the package specification be fundamentally a data-structure specification instead of fundamentally being an on-the-wire specification, we could use "file://path/to/file?format=avro" or "file+avro://path/to/file", (if avro actually supports the structure of a package file -- I haven't looked into it much).

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,350 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+                'type': 'SHA256',
+                'value': hash_file(f)
+            }
+            size = os.path.getsize(f)
+            physical_keys = [{
+                'type': PhysicalKeyType.LOCAL.name,
+                'path': os.path.abspath(f)
+            }]
+            entry = PackageEntry(physical_keys, size, hash_obj, {})
+            data[str(f)] = entry
```
dimaryaz: I think logical keys should be relative to `src_path`?

api/python/t4/snapshots.py
eode: It abstracts the package from physical keys.  It's the logical keys and data hashes that are always, always included, and ensure package reproducibility.

api/python/t4/data_transfer.py
```diff
@@ -17,6 +17,28 @@
 s3_manager = create_transfer_manager(s3_client, TransferConfig())
 
 
+def deserialize_obj(data, target):
```
dimaryaz: Can you move this back (for now at least)? It won't even work here due to missing imports.

api/python/t4/snapshots.py
meffij: I'm not sure what you mean here

api/python/t4/data_transfer.py
meffij: I finished the refactor, missing imports should be resolved now

api/python/t4/snapshots.py
dimaryaz: If I'm doing `Package.create_package('/home/dima/tmp/'), I don't want every logical key to start with `/home/dima/tmp`, right?

Also, related to this: on Windows, you'll have logical keys with backslashes...

meffij: oh I see what you're saying

api/python/t4/snapshots.py
meffij: should be `self.hash`, fixed

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,345 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        """
+        Returns the textual hash of the package manifest serialized to disk
+
+        Returns:
+            SHA256 hash of serialized manifest
+        """
+        with tempfile.NamedTemporaryFile() as tmpfile:
+            tmp_name = tmpfile.name
+            self.dump(tmpfile)
+            self_hash = hash_file(tmp_name)
```
dimaryaz: Turns out, this will have the same problem as earlier: you're now opening the temp file for the second time.

api/python/t4/snapshots.py
```diff
@@ -131,10 +139,348 @@ def download_file_from_snapshot(src, dst, snapshothash):
...
+        Returns:
+            None
+
+        Raises:
+            fail to create file
+            fail to finish write
+        """
+        writer = jsonlines.Writer(writable_file)
+        writer.write(self._meta)
+        for logical_key, entry in self._data.items():
```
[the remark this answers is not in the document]
meffij: Textual hash is probably a bad hash anyway, removed it

PR #27, opened 2018-10-18
Optimize he.put_file

dimaryaz:
Calculate ETags of the local files, get ETags of everything in the target directory, and either upload or copy files as necessary.

### DIFF
--- a/api/python/t4/data_transfer.py
+++ b/api/python/t4/data_transfer.py
@@ -151,6 +151,13 @@ def download_bytes(path, version=None):
 
 
 def _calculate_etag(file_obj):
+    """
+    Attempts to calculate a local file's ETag the way S3 does:
+    - Normal uploads: MD5 of the file
+    - Multi-part uploads: MD5 of the (binary) MD5s of the parts, dash, number of parts
+    We can't know how the file was actually uploaded - but we're assuming it was done using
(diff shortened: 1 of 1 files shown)


ocean/helium/data_transfer.py
```diff
@@ -159,6 +160,23 @@ def download_bytes(path, version=None):
     return body, meta
 
 
+def _calculate_etag(file_obj):
+    size = file_obj.stat().st_size
```
stevededalus: can you throw a docstring here that describes the two etag formats for regular and multipart?

ocean/helium/data_transfer.py
```diff
@@ -183,13 +201,33 @@ def upload_file(src_path, dest_path, meta):
...
         callback = ProgressCallback(progress)
 
         futures = []
-        for f in src_file_list:
+        for f, etag in zip(src_file_list, src_etag_list):
             real_dest_path = key + str(f.relative_to(src_root)) if (not key or key.endswith('/')) else key
-            future = s3_manager.upload(str(f), bucket, real_dest_path, extra_args, [callback])
+            existing_src = existing_etags.get(etag)
+            if existing_src is not None:
+                future = s3_manager.copy(
```
stevededalus: throw a one line comment that we are skipping uploads of preexisting files with matching hashes

PR #28, opened 2018-10-19
Implement local metadata using xattrs

dimaryaz:
xattrs are easy. Extracting metadata from the s3 transfer manager, on the other hand...
We could just make an extra HEAD request for each file - but that's inefficient. Instead, we can just patch the s3 client.

### DIFF
--- a/api/python/setup.py
+++ b/api/python/setup.py
@@ -42,6 +42,7 @@ setup(
         'ruamel.yaml<=0.15.70',
         'six>=1.10.0',
         'tqdm>=4.26.0',
+        'xattr>=0.9.6',
     ],
     extras_require={
         'tests': [
--- a/api/python/setup.py
+++ b/api/python/setup.py
@@ -42,7 +42,7 @@ setup(
         'ruamel.yaml<=0.15.70',
         'six>=1.10.0',
         'tqdm>=4.26.0',
-        'xattr>=0.9.6',
+        'xattr>=0.9.6; platform_system!="Windows"',
     ],
     extras_require={
         'tests': [
--- a/api/python/t4/api.py
+++ b/api/python/t4/api.py
@@ -16,7 +16,7 @@ from .util import (HeliumConfig, HeliumException, AWS_SEPARATOR, CONFIG_PATH,
                    write_yaml, yaml_has_comments)
 
 
-def copy(src, dest, meta=None):
+def copy(src, dest):
     """
     Copies ``src`` object from T4 to ``dest``
 
@@ -26,12 +26,8 @@ def copy(src, dest, meta=None):
     Parameters:
         src (str): a path to retrieve
         dest (str): a path to write to
-        meta (dict): optional metadata that will be assigned to the object
     """
-    all_meta = dict(
-        user_meta=meta
-    )
-    copy_file(fix_url(src), fix_url(dest), all_meta)
+    copy_file(fix_url(src), fix_url(dest))
 
 
 def put(obj, dest, meta=None):
--- a/api/python/t4/data_transfer.py
+++ b/api/python/t4/data_transfer.py
@@ -34,7 +34,6 @@ def _get_object(self, **kwargs):
     return resp
 s3_client.get_object = type(_old_get_object)(_get_object, s3_client)
 
-s3_manager = create_transfer_manager(s3_client, TransferConfig())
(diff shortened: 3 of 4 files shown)


ocean/helium/data_transfer.py
```diff
@@ -8,14 +8,27 @@
 from boto3.s3.transfer import S3Transfer, TransferConfig, create_transfer_manager
 from s3transfer.subscribers import BaseSubscriber
 from tqdm.autonotebook import tqdm
+import xattr
 
 from .util import HeliumException, split_path
 
 
 HELIUM_METADATA = 'helium'
+HELIUM_XATTR = 'user.com.quiltdata.helium'
```
dimaryaz: On Linux, it looks like the user is only allowed to set `user.*` metadata. And I'm assuming that "com.quiltdata.*" is a good naming convention (at least, that's what Dropbox does).

Can someone check if it's the same in OS X? You may be able to use `getfattr` and `setfattr` commandline tools.

akarve: I am able to set and remove on OS X with `xattr`; there is no `user.*` restriction:
```
apk-mbp-3:Desktop karve$ xattr test.txt 
com.quiltdata.helium
user.com.quiltdata.goo
user.com.quiltdata.helium
```

ocean/helium/data_transfer.py
```diff
@@ -189,6 +214,17 @@ def upload_file(src_path, dest_path, meta):
...
             real_dest_path = key + str(f.relative_to(src_root)) if (not key or key.endswith('/')) else key
+
+            try:
+                meta_bytes = xattr.getxattr(f, HELIUM_XATTR)
+                meta = json.loads(meta_bytes.decode('utf-8'))
+            except IOError:
+                # No metadata
+                meta = {}
+            except ValueError:
+                raise ValueError("Source path contains invalid metadata")
```
stevededalus: should this be 
```
except ValueError as e:
    raise ValueError("Source path contains invalid metadata") from e
```
or are you trying to mask the underlying error message?

ocean/helium/data_transfer.py
```diff
@@ -123,17 +137,29 @@ def _download_dir(bucket, prefix, dest_path):
...
 
+        metadata = {}
+        lock = Lock()
+
         futures = []
         for key, dest_file, size in tuples_list:
+            def meta_callback(resp):
+                meta = _parse_metadata(resp)
+                with lock:
+                    metadata[key] = meta
```
stevededalus: Why do you need to lock here? Are the keys not unique in tuple_list?

ocean/helium/data_transfer.py
```diff
@@ -8,14 +8,27 @@
...
 
 from .util import HeliumException, split_path
 
 
 HELIUM_METADATA = 'helium'
+HELIUM_XATTR = 'user.com.quiltdata.helium'
 
 s3_client = boto3.client('s3')
+
+_old_get_object = s3_client.get_object
```
stevededalus: Can you add a comment here about why you're patching this?

ocean/helium/data_transfer.py
dimaryaz: Interesting. Can you see what Dropbox does (assuming you have it installed)? Every file inside the Dropbox folder should have xattrs.

ocean/helium/data_transfer.py
dimaryaz: Just didn't think about it. Though `from e` won't work in Python2, if we decide to actually support it.

ocean/helium/data_transfer.py
dimaryaz: They are... But dicts aren't thread-safe?

ocean/helium/data_transfer.py
dimaryaz: Actually, I think it's done automatically; we'd need `raise ... from None` to suppress it.

ocean/helium/data_transfer.py
stevededalus: I thought all the built in types with threadsafe, but according to the google style guide I guess we should still lock. Thanks for clarifying.

Do not rely on the atomicity of built-in types.

While Python’s built-in data types such as dictionaries appear to have atomic operations, there are corner cases where they aren’t atomic (e.g. if hash or eq are implemented as Python methods) and their atomicity should not be relied upon. Neither should you rely on atomic variable assignment (since this in turn depends on dictionaries).

PR #29, opened 2018-10-19
Rename directories to remove helium references and ocean

stevededalus:
Will follow up renaming exceptions and classes in another PR.

https://app.asana.com/0/855836770921302/874855898573849/f


sdk/python/t4/api.py
```diff
@@ -374,7 +374,7 @@ def config(*autoconfig_url, **config_values):
     """Set or read the Helium configuration
 
     To retrieve the current config, call directly, without arguments:
-        >>> import helium as he
+        >>> import t4 as he
         >>> he.config()
```
dimaryaz: We should probably call it `t4` rather than `he` now?

stevededalus: Yeah, going to do a follow up rename of all these and the Exceptions names and docs.

akarve:
i suspect people looking for a "way to use T4 in Python" would find `api/` more easily, but i'm not religous about this so pick the one you think works best

stevededalus:
Ok, switched to API, PTAL

PR #31, opened 2018-10-24
Add some integration tests for packages

stevededalus:


api/python/tests/integration/test_packages.py
```diff
@@ -0,0 +1,57 @@
...
+def mock_make_api_call(operation_name):
+    """ Mock boto3's AWS API Calls for testing. """
+    if operation_name == 'GetObject':
+        parsed_response = {'Body': {'foo'}}
+        return parsed_response
+    raise NotImplementedError
+
+def test_read_manifest(tmpdir):
+    """ Verify reading serialized manifest from disk. """
+    pkg = Package.load(open(LOCAL_MANIFEST))
```
dimaryaz: Use `with open(...) as fd:`

api/python/tests/integration/test_packages.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+    with pytest.raises(NotImplementedError):
+        pkg.get('foo')
+
+    out_path = os.path.join(tmpdir, 'new_manifest.jsonl')
+    pkg.dump(open(out_path, "w"))
```
dimaryaz: Same here. Here it actually matters cause the data might not get written until the file object is garbage collected at some arbitrary time in the future.

api/python/tests/integration/test_packages.py
```diff
@@ -0,0 +1,57 @@
...
+    
+    # Insepct the jsonl to verify everything is maintained, i.e.
+    # that load/dump results in an equivalent set.
+    original_set = list(jsonlines.Reader(open(LOCAL_MANIFEST)))
```
dimaryaz: Wouldn't it be easier to just `Package.load` the second file and compare the contents of the packages? That would verify that write/read works without depending on the internals of it.

api/python/tests/integration/test_packages.py
```diff
@@ -0,0 +1,57 @@
...
+    written_set = list(jsonlines.Reader(open(out_path)))
+    assert len(original_set) == len(written_set)
+    assert sorted(original_set, key=lambda k: k.get('logical_key','manifest')) \
+        == sorted(written_set, key=lambda k: k.get('logical_key','manifest'))
+
+def test_materialize_from_remote():
+    """ Verify loading data and mainfest transforms from S3. """
+    with patch('botocore.client.BaseClient._make_api_call', new=mock_make_api_call):
```
dimaryaz: It's fine for something simple like this... But in general, look into `botocore.stub.Stubber`. See the old registry tests for examples: https://github.com/quiltdata/quilt/blob/master/registry/tests/utils.py

api/python/tests/integration/test_packages.py
stevededalus: Added a todo for now, needs the __eq__ to make this easier to work with.

api/python/tests/integration/test_packages.py
stevededalus: Will take a look at that in a follow up.

api/python/tests/integration/test_packages.py
dimaryaz: Ok. Can you use `with open` here and below, though?

PR #32, opened 2018-10-24
First draft of top_hash

meffij:
TODO: make sure this has the properties we want it to have (changes when any contents change except physical key, doesn't care about serialization / dict order / etc)

### DIFF
--- a/api/python/t4/snapshots.py
+++ b/api/python/t4/snapshots.py
@@ -457,7 +457,6 @@ class Package(object):
         hashable_meta.pop('top_hash', None)
         top_meta = json.dumps(hashable_meta, sort_keys=True, separators=(',', ':'))
         top_hash.update(top_meta.encode('utf-8'))
-        current = 0
         for logical_key, entry in sorted(list(self._data.items())):
             entry_dict = entry.as_dict()
             entry_dict['logical_key'] = logical_key
--- a/api/python/t4/snapshots.py
+++ b/api/python/t4/snapshots.py
@@ -166,40 +166,6 @@ def bytes_to_int(input_bytes):
         result = result * 256 + int(byte)
     return result
 
-HASH_MAX = 2 ** 256
-
-def sum_hash(current, add_bytes):
-    """
-    Sums hashes for order-independence.
-
-    Args:
-        current(int)
-        add_bytes(bytes)
-
-    Returns:
-        New total
-    """
-    total = current + bytes_to_int(add_bytes)
-    total = total % HASH_MAX
-    return total
-
-def hash_object(obj):
-    hasher = hashlib.sha256()
-    if isinstance(obj, dict):
-        current = 0
-        for key, value in obj.items():
-            item_hash = hashlib.sha256()
-            item_hash.update(hash_object(key).encode('utf-8'))
-            item_hash.update(hash_object(value).encode('utf-8'))
-            current = sum_hash(current, item_hash.digest())
-        hasher.update(str(current).encode('utf-8'))
-    elif isinstance(obj, list):
-        for item in obj:
-            hasher.update(hash_object(item).encode('utf-8'))
-    else:
-        hasher.update(str(obj).encode('utf-8'))
-    return hasher.hexdigest()
-
 def dereference_physical_key(physical_key):
     ty = physical_key['type']
     if ty == PhysicalKeyType.LOCAL.name:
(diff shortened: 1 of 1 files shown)


api/python/t4/snapshots.py
```diff
@@ -160,6 +160,46 @@ def hash_file(readable_file):
...
+        add_bytes(bytes)
+
+    Returns:
+        New total
+    """
+    total = current + bytes_to_int(add_bytes)
+    total = total % HASH_MAX
+    return total
+
+def hash_object(obj):
```
eode: `PackageEntry.hash()`?

api/python/t4/snapshots.py
```diff
@@ -308,6 +348,7 @@ def create_package(path):
 
             size = os.path.getsize(f)
             physical_keys = [{
+                'version': '0.0.1',
```
stevededalus: Why do we need version on the physical_keys?

api/python/t4/snapshots.py
meffij: The top hash of a Package intentionally excludes the physical keys, which is not necessarily a property of the PackageEntry. Also, PackageEntry objects are mutable (which I don't have a convenient way of preventing) so they don't have a stable hash.

api/python/t4/snapshots.py
meffij: It was part of the original proposed schema. I will drop it. We also decided to drop type from the physical keys for now

api/python/t4/snapshots.py
```diff
@@ -426,7 +468,23 @@ def _top_hash(self):
...
             None
         """
-        raise NotImplementedError
+        top_hash = hashlib.sha256()
+        hashable_meta = copy.deepcopy(self._meta)
+        hashable_meta.pop('top_hash', None)
+        top_meta_hash = hash_object(hashable_meta)
+        top_hash.update(top_meta_hash.encode('utf-8'))
+        current = 0
+        for lk, entry in self._data.items():
```
stevededalus: s/lk/logical_key for readability

api/python/t4/snapshots.py
```diff
@@ -160,6 +160,46 @@ def hash_file(readable_file):
...
+
+    Args:
+        current(int)
+        add_bytes(bytes)
+
+    Returns:
+        New total
+    """
+    total = current + bytes_to_int(add_bytes)
+    total = total % HASH_MAX
```
stevededalus: Why not just XOR the bit arrays instead of having all this arithmetic. They're both insecure ways of combining the hashes, so if you looking for an MVP might as well do that.

api/python/t4/snapshots.py
```diff
@@ -160,6 +160,46 @@ def hash_file(readable_file):
 
     return hasher.hexdigest()
 
+def bytes_to_int(input_bytes):
```
stevededalus: I think this is the same as `int.from_bytes()`, but worth double checking

api/python/t4/snapshots.py
```diff
@@ -426,7 +428,23 @@ def _top_hash(self):
         Returns:
             None
         """
-        raise NotImplementedError
+        top_hash = hashlib.sha256()
+        hashable_meta = copy.deepcopy(self._meta)
+        hashable_meta.pop('top_hash', None)
```
dimaryaz: Not particularly important... But it seems like `top_hash` shouldn't even be a part of the metadata. Probably don't want to save it in `dump`, either.

meffij: I think you're right; removing it from dump

meffij: removed

api/python/t4/snapshots.py
stevededalus: look like this is still here. Leave type in for now, Dima's PRs have coded against it. We can rip out tomorrow later

stevededalus: actually, we need version. Version ids are needed by boto, it doesn't handle query parameters correctly in the url so we need to explicitly

dimaryaz: I don't think this one was supposed to be the S3 version ID. In any case, I'll rebase my PR and add whatever is necessary.

meffij: The version was always the physical_key schema version, not the s3 version or anything. I should have probably named it schema_version from the get-go

PR #33, opened 2018-10-25
Make set do more things

meffij:


api/python/t4/snapshots.py
```diff
@@ -216,6 +216,21 @@ def as_dict(self):
...
+                'type': 'SHA256',
+                'value': hash_file(file_to_hash)
+            }
+
+        size = os.path.getsize(path)
+        physical_keys = [{
+            'type': PhysicalKeyType.LOCAL.name,
+            'path': os.path.abspath(path)
+        }]
+        return PackageEntry(physical_keys, size, hash_obj, {})
```
stevededalus: Add a todo to ingest metadata here when https://github.com/quiltdata/t4/pull/28 is ready

api/python/t4/snapshots.py
```diff
@@ -400,9 +404,22 @@ def update(self, logical_key, entry):
         Returns:
             A new package
         """
+        if 'meta' in kwargs and entry is None:
+            return self._update_meta(logical_key, kwargs['meta'])
+
+        pkg = self._clone()
+        if isinstance(entry, str):
```
stevededalus: Update the docstring arg type to reflect that it can be a string now.

api/python/t4/snapshots.py
```diff
@@ -400,9 +404,22 @@ def update(self, logical_key, entry):
         Returns:
             A new package
         """
+        if 'meta' in kwargs and entry is None:
```
stevededalus: Just make meta a named optional keyword arg in the signature

api/python/t4/snapshots.py
```diff
@@ -400,9 +404,22 @@ def update(self, logical_key, entry):
...
+            return self._update_meta(logical_key, kwargs['meta'])
+
```
stevededalus: raise an exception if entry and meta are both missing?

api/python/t4/snapshots.py
meffij: done

PR #34, opened 2018-10-25
Implement Package.get and Package.copy

dimaryaz:


api/python/t4/snapshots.py
```diff
@@ -147,8 +182,8 @@ def download_bytes_from_snapshot(src, snapshothash):
     return download_bytes(src, version=obj_rec['VersionId'])
 
 class PhysicalKeyType(Enum):
```
akarve: Do we actually need `PhysicalKeyType` at all if we're using URIs?

akarve: I think we want an enum of URI schemes instead? maybe that is for a separate PR.

api/python/t4/snapshots.py
```diff
@@ -333,18 +378,28 @@ def get(self, logical_key):
...
-            # TODO: dispatch on target to deserialize
-            raise NotImplementedError
 
-        raise NotImplementedError
+        return deserialize_obj(data, target), entry.meta.get('user_meta')
 
-    def get_files(self, logical_key, path):
+    def copy(self, logical_key, dest):
         """
         Gets objects from logical_key inside the package and saves them to path.
```
akarve: "saves them to dest"?

api/python/t4/snapshots.py
```diff
@@ -333,18 +378,28 @@ def get(self, logical_key):
...
         # TODO: verify hash
-        if 'target' in entry.meta:
-            # TODO: dispatch on target to deserialize
-            raise NotImplementedError
 
-        raise NotImplementedError
+        return deserialize_obj(data, target), entry.meta.get('user_meta')
 
-    def get_files(self, logical_key, path):
+    def copy(self, logical_key, dest):
```
akarve: does `api.py` need to change with this? i.e. are we ready to make `get_file` and `put_file` into `copy`? OK to punt on this.

api/python/t4/snapshots.py
dimaryaz: Probably not

api/python/t4/snapshots.py
dimaryaz: Yeah, we probably want to do that - but wanted to make sure that URIs actually make sense

api/python/t4/snapshots.py
```diff
@@ -45,6 +47,39 @@ def _parse_snapshot_path(snapshot_file_key):
...
+            # TODO: metadata
+            shutil.copyfile(src_url.path, dest_url.path)
+        elif dest_url.scheme == 's3':
+            upload_file(src_url.path, dest_url.netloc + dest_url.path, meta)
+        else:
+            raise NotImplementedError
+    elif src_url.scheme == 's3':
+        if dest_url.scheme == 'file':
+            # TODO: metadata
+            download_file(src_url.netloc + src_url.path, dest_url.path)
```
stevededalus: I think we need params here so we can download the versioned files?

api/python/t4/snapshots.py
```diff
@@ -45,6 +47,39 @@ def _parse_snapshot_path(snapshot_file_key):
...
+        elif dest_url.scheme == 's3':
+            copy_object(src_url.netloc + src_url.path, dest_url.netloc + dest_url.path, meta)
```
stevededalus: maybe params here too?

api/python/t4/snapshots.py
```diff
@@ -160,12 +195,22 @@ def hash_file(readable_file):
...
+        if url.scheme != 'file':
+            raise ValueError('Unexpected scheme: %r' % url.scheme)
+        with open(url.path, 'rb') as fd:
+            return fd.read()
+    elif ty == PhysicalKeyType.S3:
+        if url.scheme != 's3':
+            raise ValueError('Unexpected scheme: %r' % url.scheme)
+        return download_bytes(url.netloc + url.path)[0]
+    else:
+        assert False
```
stevededalus: Raise an exception with unsupport physical key type (maybe "unsupported storage" is more user friendly)?

api/python/t4/snapshots.py
```diff
@@ -333,18 +378,28 @@ def get(self, logical_key):
...
+
+        data = read_physical_key(physical_key)
+
         # TODO: verify hash
-        if 'target' in entry.meta:
-            # TODO: dispatch on target to deserialize
-            raise NotImplementedError
 
-        raise NotImplementedError
+        return deserialize_obj(data, target), entry.meta.get('user_meta')
```
stevededalus: update the return type in the docstring

PR #35, opened 2018-10-26
First sketch of package constructor

meffij:


api/python/t4/snapshots.py
```diff
@@ -167,12 +167,13 @@ def dereference_physical_key(physical_key):
...
     raise NotImplementedError
 
-def get_package_registry(path):
-    """ Returns the package registry for a given path """
+def get_package_registry_root(path=''):
+    """ Returns the package registry root for a given path """
     if path.startswith('s3://'):
         bucket = path[5:].partition('/')[0]
-        return "s3://{}/.quilt/packages/".format(bucket)
+        return "s3://{}".format(bucket)
```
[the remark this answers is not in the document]
meffij: I think registries should be defined as the path up to `.quilt/...`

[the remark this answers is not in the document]
meffij: That makes it easier for users to specify by a couple characters, I don't see any other tiebreakers

api/python/t4/snapshots.py
```diff
@@ -219,18 +221,53 @@ def as_dict(self):
...
+            registry(string): location of registry to load package from
+        """
+        if name is None and pkg_hash is None and registry is None:
+            self._data = {}
+            self._meta = {'version': 'v0'}
+            return
+
+        if registry is None:
+            # default to local registry
+            registry = get_package_registry_root()
```
stevededalus: should be able rebase/merge and get my changes for local registry here

api/python/t4/packages.py
```diff
@@ -85,13 +86,13 @@ def read_physical_key(physical_key):
         raise NotImplementedError
 
 
-def get_package_registry(path):
-    """ Returns the package registry for a given path """
+def get_package_registry(path=''):
+    """ Returns the package registry root for a given path """
```
stevededalus: Looks like it returns a string in one case and a path in the other, can you update? probably has to be a file uri string for the local package instead of a Path object

api/python/t4/packages.py
```diff
@@ -158,18 +159,69 @@ def _clone(self):
...
+            self._meta = {'version': 'v0'}
+            return
+
+        if registry is None:
+            # default to local registry
+            registry = get_package_registry_root()
+
+        if pkg_hash is not None:
+            # if hash is specified, name doesn't matter
+            pkg_path = registry + 'packages/{}'.format(pkg_hash)
```
stevededalus: maybe `'{}packages/{}'.format(registry, pkg_hash)` so you aren't mixing concat and format?

api/python/t4/packages.py
```diff
@@ -158,18 +159,69 @@ def _clone(self):
...
+                with open(pkg_path) as pkg_file:
+                    pkg = self.load(pkg_file)
+            elif pkg_path.startswith('s3://'):
+                body, _ = download_bytes(pkg_path)
+                pkg = self.load(io.BytesIO(body))
+
+            self = pkg._clone()
+            return
+
+        pkg_path = registry + 'named_packages/{}/'.format(name)
```
stevededalus: `quote(name)`

api/python/t4/packages.py
```diff
@@ -158,18 +159,69 @@ def _clone(self):
...
+        latest = pkg_path + 'latest'
+        if latest.startswith('file:///'):
+            with open(latest) as latest_file:
+                latest_hash = latest_file.read()
```
stevededalus: might need to strip a trailing line break in the hash file

api/python/t4/packages.py
```diff
@@ -158,18 +159,69 @@ def _clone(self):
...
+        elif latest.startswith('s3://'):
+            no_scheme_path = latest[5:]
+            latest_bytes = download_bytes(no_scheme_path)
+            latest_hash = latest_bytes.decode('utf-8')
+        else:
+            raise NotImplementedError
+
+        latest_path = registry + 'packages/{}'.format(latest_hash)
+        if latest_path.startswith('file:///'):
```
stevededalus: Can this block and the block at 184-190 be the same function?

stevededalus:
Add a test to the `integration/test_packages.py` file

ResidentMario:
screen shot 2018-10-26 at 5 10 29 pm

@meffij Your code might be creating shadow file structure, due to a path with an extra forward slash (`/`) in a path somewhere.

api/python/t4/packages.py
```diff
@@ -85,13 +86,13 @@ def read_physical_key(physical_key):
...
 
 
-def get_package_registry(path):
-    """ Returns the package registry for a given path """
+def get_package_registry(path=''):
+    """ Returns the package registry root for a given path """
     if path.startswith('s3://'):
         bucket = path[5:].partition('/')[0]
-        return "s3://{}/.quilt".format(bucket)
+        return "s3://{}/.quilt/".format(bucket)
```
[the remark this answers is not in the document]
meffij: OK will change

api/python/t4/packages.py
```diff
@@ -155,21 +156,80 @@ def _clone(self):
...
+        if pkg_hash is not None:
+            # if hash is specified, name doesn't matter
+            pkg_path = '{}/packages/{}'.format(registry, pkg_hash)
+            # TODO replace open with something that supports both local and s3
+            self = self._from_path(pkg_path)
+            return
+
+        pkg_path = '{}/named_packages/{}/'.format(registry, quote(name))
+        latest = pkg_path + 'latest'
+        if latest.startswith('file:///'):
```
dimaryaz: I just merged my PR with some URI code. Can you update this to parse the URL similar to `copy_file` in `data_transfer.py`? Specifically, use `url2pathname` and `parse_s3_url`.

I'm still trying to come up with some abstraction to avoid `if``/`elif` everywhere we deal with S3 and files... Not sure how to do it.

api/python/t4/packages.py
```diff
@@ -155,21 +156,80 @@ def _clone(self):
...
+        else:
+            raise NotImplementedError
+
+        latest_hash = latest_hash.strip()
+        latest_path = '{}/packages/{}'.format(registry, quote(latest_hash))
+        self = self._from_path(latest_path)
+
+    @staticmethod
+    def _from_path(path):
+        if path.startswith('file:///'):
```
dimaryaz: Same here.

api/python/t4/packages.py
meffij: As far as the if/elif chains go, I think factoring out interfaces like `StorageProvider`, etc will help us use dynamic dispatch instead of if/elif

api/python/t4/packages.py
```diff
@@ -113,21 +114,77 @@ def _clone(self):
...
+
+    @staticmethod
+    def _from_path(uri):
+        """ Takes a URI and returns a package loaded from that URI """
+        src_url = urlparse(uri)
+        if src_url.scheme == 'file':
+            with open(unquote(src_url.path)) as open_file:
+                pkg = self.load(open_file)
+        elif src_url.scheme == 's3':
+            body, _ = download_bytes(src_url.geturl())
```
dimaryaz: Does this work, though?.. `download_bytes` doesn't accept URLs (yet)

api/python/t4/packages.py
```diff
@@ -113,21 +114,77 @@ def _clone(self):
...
+        latest_hash = latest_hash.strip()
+        latest_path = '{}/packages/{}'.format(registry, quote(latest_hash))
+        self = self._from_path(latest_path)
+
+    @staticmethod
+    def _from_path(uri):
+        """ Takes a URI and returns a package loaded from that URI """
+        src_url = urlparse(uri)
+        if src_url.scheme == 'file':
+            with open(unquote(src_url.path)) as open_file:
```
dimaryaz: `url2pathname`, not `unquote`. Though I'm still figuring out how to do this properly :-\

PR #36, opened 2018-10-26
Implement build that registers a new local package

stevededalus:
Implement build that registers a new local package
- Places a latests and timestamp pointer in named_packages/<PACKAGE_NAME>
- Places a manifest in packages/<TOP_HASH>

### DIFF
--- a/api/python/t4/snapshots.py
+++ b/api/python/t4/snapshots.py
@@ -464,7 +464,8 @@ class Package(object):
         Returns:
             None
         """
-        with open(os.path.join(get_local_package_registry(), "packages", name), "w") as fh:
+        hash_string = self.top_hash()["value"]
+        with open(os.path.join(get_local_package_registry(), "packages", hash_string), "w") as fh:
             self.dump(fh)
 
         # Build the package directory if necessary.
@@ -472,10 +473,11 @@ class Package(object):
         Path(named_path).mkdir(parents=True, exist_ok=True)
         # todo: use a float to string formater instead of double casting
         with open(os.path.join(named_path, str(int(time.time()))), "w") as fh:
-            fh.write(self.top_hash())
-        # todo: symlink
+            fh.write(self.top_hash()["value"])
+        # todo: symlink when local
         with open(os.path.join(named_path, "latest"), "w") as fh:
-            fh.write(self.top_hash())
+            fh.write(self.top_hash()["value"])
+        return hash_string
 
     def dump(self, writable_file):
         """
--- a/api/python/t4/snapshots.py
+++ b/api/python/t4/snapshots.py
@@ -1,18 +1,23 @@
+import appdirs
 import copy
 from enum import Enum
 import hashlib
 import json
 import pathlib
 import os
+
 import shutil
+import time 
 from urllib.parse import parse_qs, unquote, urlparse
 from urllib.request import url2pathname
 
 import boto3
 import jsonlines
 
+from pathlib import Path
 from .data_transfer import (copy_object, deserialize_obj, download_bytes, download_file, list_objects,
                             list_object_versions, upload_bytes, upload_file, TargetType)
+
 from .exceptions import PackageException
 from .util import HeliumException, split_path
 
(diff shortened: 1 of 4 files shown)


api/python/tests/integration/test_packages.py
```diff
@@ -18,6 +19,49 @@ def mock_make_api_call(operation_name):
...
+
+    # Build a new package into the local registry.
+    new_pkg = new_pkg.set('foo', test_file)
+    top_hash = new_pkg.build("Test")
+
+    # Verify manifest is registered by hash.
+    out_path = os.path.join(appdirs.user_data_dir("quilt"), "packages", top_hash)
+    with open(out_path) as fd:
+        pkg = Package.load(fd)
+        assert "file://" + test_file \
```
dimaryaz: Well, it's a URI, so it's not that simple... You technically need to quote it. If you dare to try it on Windows, it's even crazier:
```
E           AssertionError: assert 'file://C:\\u...t_build0\\bar' == 'file:///C:/us...st_build0/bar'
E             - file://C:\users\dima\Temp\pytest-of-dima\pytest-1\test_build0\bar
E             ?          ^     ^    ^    ^              ^        ^           ^
E             + file:///C:/users/dima/Temp/pytest-of-dima/pytest-1/test_build0/bar
E             ?        +  ^     ^    ^    ^              ^        ^           ^
```
But if you use `Path`, you should be able to just call `test_file.resolve().as_uri()`.

Alternatively, just call `pkg.get(...)` and not worry about the exact URL value.

api/python/tests/integration/test_packages.py
```diff
@@ -18,6 +19,49 @@ def mock_make_api_call(operation_name):
...
 
+def test_build(tmpdir):
+    """Verify that build dumps the manifest to appdirs directory."""
+    # Patch appdirs to use a temp dir.
+    appdirs.user_data_dir = lambda x: os.path.join(tmpdir, x)
+
+    new_pkg = Package()
+
+    # Create a dummy file to add to the package.
+    test_file = os.path.join(tmpdir, 'bar')
```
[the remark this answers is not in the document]
stevededalus: I need the full path for the physical key assertion below. I guess I could do 
```
with open('bar', 'w') as fd:
    fd.write("assa")
    test_file = os.path.abspath(fd)
```
but that doesn't seem much more readable.

api/python/t4/snapshots.py
```diff
@@ -441,6 +453,33 @@ def get_meta(self, logical_key):
...
+        hash_string = self.top_hash()["value"]
+        with open(os.path.join(get_local_package_registry(), "packages", hash_string), "w") as fh:
+            self.dump(fh)
+
+        if name:
+            # Build the package directory if necessary.
+            named_path = os.path.join(get_local_package_registry(), "named_packages", name)
+            Path(named_path).mkdir(parents=True, exist_ok=True)
+            # todo: use a float to string formater instead of double casting
+            with open(os.path.join(named_path, str(int(time.time()))), "w") as fh:
```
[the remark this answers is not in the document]
stevededalus: Yeah, the file names are an implementation and aren't used by anything, so we can change them to something else. Let's leave for now.

api/python/t4/snapshots.py
```diff
@@ -441,6 +451,33 @@ def get_meta(self, logical_key):
...
+            the top hash as a string
+        """
+        hash_string = self.top_hash()["value"]
+        with open(get_local_package_registry() / "packages" / hash_string, "w") as fh:
+            self.dump(fh)
+
+        if name:
+            # Build the package directory if necessary.
+            named_path = get_local_package_registry() / "named_packages" / name
+            Path(named_path).mkdir(parents=True, exist_ok=True)
```
dimaryaz: Don't need the `Path` here - it's already a `Path`

api/python/tests/integration/test_packages.py
```diff
@@ -18,6 +20,44 @@ def mock_make_api_call(operation_name):
         return parsed_response
     raise NotImplementedError
 
+@patch('appdirs.user_data_dir', side_effect=(lambda x,y: os.path.join('test_appdir', x)))
```
dimaryaz: `side_effect=` doesn't seem right to me... Just this should work:
```
@patch('appdirs.user_data_dir', lambda x,y: os.path.join('test_appdir', x))
```
But doesn't really matter, I guess

PR #37, opened 2018-10-26
Add a push implementation and  test

stevededalus:


api/python/t4/snapshots.py
```diff
@@ -554,18 +564,63 @@ def top_hash(self):
...
+            # todo: handle where to put data for unnamed remote packages
+            raise NotImplementedError
+        pkg = self._materialize('{}/{}'.format(path.strip("/"), name))
+
+        with tempfile.NamedTemporaryFile() as manifest:
+            pkg.dump(manifest)
+            manifest.flush()
+            _copy_file(
+                _fix_url(manifest.name),
+                get_package_registry(path) + "/packages/" + pkg.top_hash()["value"],
```
dimaryaz: I wonder if we should get rid of `["value"]` and just have one single correct hash...

api/python/t4/snapshots.py
```diff
@@ -554,18 +564,63 @@ def top_hash(self):
...
+        if not name:
+            # todo: handle where to put data for unnamed remote packages
+            raise NotImplementedError
+        pkg = self._materialize('{}/{}'.format(path.strip("/"), name))
+
+        with tempfile.NamedTemporaryFile() as manifest:
+            pkg.dump(manifest)
+            manifest.flush()
+            _copy_file(
+                _fix_url(manifest.name),
```
dimaryaz: I meant it to be used on user input where we want to support both URIs and file paths - and possibly use heuristics to guess which one it is. When we know it's a file path, I'd rather use `Path(manifest.name).resolve().as_uri()` - or change the other helpers to make it less verbose.

api/python/t4/snapshots.py
```diff
@@ -575,8 +630,23 @@ def materialize(self, path, name=None):
...
-        self.get_files(path)
-        self.dump(get_package_registry(path) + name)
+        pkg = self._clone()
+        for logical_key, entry in self._data.items():
+            # Copy the datafiles in the package.
+            if path.startswith('s3://'):
+                new_physical_key = path + "/" + logical_key
+                new_type = PhysicalKeyType.S3.name
+            else:
+                new_physical_key = os.path.join(path, logical_key)
```
dimaryaz: These are supposed to be URIs, not paths... Need to parse `path` and re-build it - or just pass the parsed URI into this method. Or, if we know `path` is a URI, then something like `path + '/' + quote(logical_key)` would be correct.

api/python/t4/snapshots.py
stevededalus: Agreed, added a task to track refactoring this
https://app.asana.com/0/855836770921302/884535477618849/f

api/python/t4/snapshots.py
stevededalus: Swapped to use resolve as uri. Leaving as a string to path instead of uri for now because this is supposed to help with downloading all the contents of a package, so nice to be able to use a relative path. This will end up getting refactored though once we have a stable user facing way to call materialize.

api/python/t4/snapshots.py
```diff
@@ -637,7 +637,7 @@ def _materialize(self, path):
                 new_physical_key = path + "/" + logical_key
                 new_type = PhysicalKeyType.S3.name
             else:
-                new_physical_key = os.path.join(path, logical_key)
+                new_physical_key = os.path.join(path, logical_key).resolve().as_uri()
```
dimaryaz: This won't work, though: `os.path.join` gives you a string, not a Path. In any case... you should have exactly the same logic in `if` and in `else`, since it's a URI in both cases.

api/python/t4/snapshots.py
```diff
@@ -591,18 +600,64 @@ def top_hash(self):
...
+        Returns:
+            A new package that points to the copied objects
+        """
+        dest = _fix_url(path)
+        if not dest.startswith('s3://'):
+            raise NotImplementedError
+        if not name:
+            # todo: handle where to put data for unnamed remote packages
+            raise NotImplementedError
+        pkg = self._materialize('{}/{}'.format(dest.strip("/"), name))
```
dimaryaz: Need to quote name here, now that this is a URI

api/python/t4/snapshots.py
```diff
@@ -612,8 +667,25 @@ def materialize(self, path, name=None):
             fail to put bytes
             fail to put package to registry
         """
-        raise NotImplementedError
-        if name is None:
-            raise NotImplementedError
-        self.get_files(path)
-        self.dump(get_package_registry(path) + name)
+        pkg = self._clone()
+        dest = _fix_url(path)
```
dimaryaz: Don't need it here (since this isn't a public API). Maybe assert?

PR #46, opened 2018-10-27
Replace get_file and put_file with copy

dimaryaz:


api/python/t4/util.py
```diff
@@ -70,6 +71,29 @@ def split_path(path, require_subpath=False):
     return result
 
 
+def fix_url(url):
```
stevededalus: thanks for moving these here

PR #48, opened 2018-10-30
Yet another attempt at making URIs work properly

dimaryaz:
Handle weird corner cases on Windows.

### DIFF
--- a/api/python/t4/util.py
+++ b/api/python/t4/util.py
@@ -102,7 +102,7 @@ def parse_file_url(file_url):
     if file_url.scheme != 'file':
         raise ValueError("Invalid file URI")
     path = url2pathname(file_url.path)
-    if file_url.netloc:
+    if file_url.netloc not in ('', 'localhost'):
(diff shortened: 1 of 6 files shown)


api/python/t4/util.py
```diff
@@ -94,6 +98,17 @@ def parse_s3_url(s3_url):
     return bucket, path, version_id
 
 
+def parse_file_url(file_url):
+    if file_url.scheme != 'file':
+        raise ValueError("Invalid file URI")
+    path = url2pathname(file_url.path)
+    if file_url.netloc:
+        # Windows file share
+        # TODO: Can't do anything useful on non-Windows... Return an error?
```
meffij: The file protocol technically accepts netlocs like `localhost`... Maybe there is sane behavior on posix too?

dimaryaz: Yeah, just found out about it after reading the wikipedia page. `localhost` is easy. What about all the others, though - error?

PR #49, opened 2018-10-30
list_packages implementation

stevededalus:


api/python/t4/api.py
```diff
@@ -60,6 +64,34 @@ def ls(path, recursive=False):
     return results
 
 
+def list_packages(registry=None):
+    """ Lists Packages in the registry.
+
+    Returns a list of all named packages in a registry.
+    If the registry in None, default to the local registry.
```
kevinemoore: "is None"?

api/python/t4/api.py
```diff
@@ -60,6 +64,34 @@ def ls(path, recursive=False):
...
+        registry(string): location of registry to load package from.
+
+    Returns:
+        A list of strings containing the names of the packages        
+    """
+    if not registry:
+        registry = ''
+    registry = get_package_registry(fix_url(registry)).strip("/") + '/named_packages'
+
+    registry_url = urlparse(registry)
```
kevinemoore: Is this the only function that needs these primitives (e.g., switching based on the URL type?). If not, maybe it's worth creating a wrapper for registries?

stevededalus: Yeah, Dima is working on this right now, partially related to https://github.com/quiltdata/t4/pull/48

PR #53, opened 2018-10-30
Fix __init__

meffij:


api/python/t4/packages.py
```diff
@@ -158,7 +159,8 @@ def __init__(self, name=None, pkg_hash=None, registry=None):
 
         latest_hash = latest_hash.strip()
         latest_path = '{}/packages/{}'.format(registry, quote(latest_hash))
-        self = self._from_path(latest_path)
+        pkg = self._from_path(latest_path)
+        self._set_state(pkg._data, pkg._meta)
```
stevededalus: can you add a comment about why you need to call `set_state` here?

PR #54, opened 2018-10-30
Unset top_hash when returning modified package

meffij:
If we don't do this, then the new top hash will never be calculated (since it's cached in _meta) and packages will be hanging around with incorrect top hashes.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -262,6 +262,8 @@ class Package(object):
             logical_key = prefix + f.relative_to(src_path).as_posix()
             # TODO: Warn if overwritting a logical key?
             pkg = pkg.set(logical_key, entry)
+
+        pkg._unset_tophash()
         return pkg
 
     def get(self, logical_key):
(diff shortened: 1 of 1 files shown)


api/python/t4/packages.py
```diff
@@ -262,6 +262,8 @@ def capture(self, path, prefix=None):
             logical_key = prefix + f.relative_to(src_path).as_posix()
             # TODO: Warn if overwritting a logical key?
             pkg = pkg.set(logical_key, entry)
+
+        pkg._unset_tophash()
```
stevededalus: Can you add an inline comment about why you have to do this and also explain it in the PR description?

api/python/t4/packages.py
```diff
@@ -483,6 +489,9 @@ def _top_hash(self):
             'value': top_hash.hexdigest()
         }
 
+    def _unset_tophash(self):
```
stevededalus: Doc string explaining why its necessary

api/python/t4/packages.py
```diff
@@ -402,6 +405,8 @@ def update(self, new_keys_dict, meta=None):
         pkg = self._clone()
         for logical_key, entry in new_keys_dict.items():
             pkg = pkg.set(logical_key, entry, meta)
+        # modified package should have new top hash
+        pkg._unset_tophash()
```
stevededalus: Isn't this unset by the `pkg.set` call above?

meffij: it sure is, will fix

PR #55, opened 2018-10-30
Fix s3 package

meffij:


ResidentMario:
There's a CI failure however that look to be due to this change impacting one of the tests.

PR #56, opened 2018-10-30
Infer .quilt on s3 bucket registries

meffij:


api/python/t4/packages.py
```diff
@@ -136,6 +136,14 @@ def __init__(self, name=None, pkg_hash=None, registry=None):
             # default to local registry
             registry = get_package_registry()
 
+        parsed_registry = urlparse(registry)
+        if parsed_registry.scheme == 's3' and parsed_registry.path in ['', '/']:
```
stevededalus: shouldn't this logic be in the get_registry function so its consistent with push and build?

meffij: no, since this is an override of a user-specified registry rather than an inferred registry from a path

stevededalus: but isn't this what `get_registry(path)` does? It takes a path determines if its s3 and the returns the bucket + `/.quilt`

stevededalus: list_packages also uses a user supplied registry.

meffij: I think you're right, we may need to split the code eventually but for now we can use `get_registry`

PR #57, opened 2018-10-30
Fix handling of prefix response from S3 list_obj api

stevededalus:


api/python/t4/api.py
```diff
@@ -86,10 +86,12 @@ def list_packages(registry=None):
     elif registry_url.scheme == 's3':
         src_bucket, src_path, _ = parse_s3_url(registry_url)
         prefixes, _ = list_objects(src_bucket + '/' + src_path, recursive=False)
+        print(prefixes)
```
ResidentMario: Leftover `print` statement.

stevededalus: Like a hawk, thanks.

ResidentMario:
Works on my machine, LGTM.

PR #59, opened 2018-10-30
Suppress exception

meffij:
boto3 throws when we try to copy an object in s3 to itself without changing the metadata, storage class, etc. Suppress this exception because `copy_object` should work even when its source and destination are the same.

We can't tell whether a copy will be a no-op since we don't know if the provided `meta` argument matches the metadata on the source object.

### DIFF
--- a/api/python/t4/data_transfer.py
+++ b/api/python/t4/data_transfer.py
@@ -8,6 +8,7 @@ from threading import Lock
 from urllib.parse import urlparse
 from urllib.request import url2pathname
 
+from botocore.exceptions import ClientError
 import boto3
 from boto3.s3.transfer import TransferConfig, create_transfer_manager
 from s3transfer.subscribers import BaseSubscriber
@@ -346,6 +347,7 @@ def delete_object(path):
         s3_client.head_object(Bucket=bucket, Key=key)  # Make sure it exists
         s3_client.delete_object(Bucket=bucket, Key=key)  # Actually delete it
 
+NO_OP_COPY_ERROR_MESSAGE = "An error occurred (InvalidRequest) when calling the CopyObject operation: This copy request is illegal because it is trying to copy an object to itself without changing the object's metadata, storage class, website redirect location or encryption attributes."
 
 def copy_object(src, dest, meta, version=None):
     src_bucket, src_key = split_path(src, require_subpath=True)
@@ -359,13 +361,18 @@ def copy_object(src, dest, meta, version=None):
             VersionId=version
         )
 
-    s3_client.copy_object(
-        CopySource=src_params,
-        Bucket=dest_bucket,
-        Key=dest_key,
-        Metadata={HELIUM_METADATA: json.dumps(meta)}
-    )
-
+    try:
+        s3_client.copy_object(
+            CopySource=src_params,
+            Bucket=dest_bucket,
+            Key=dest_key,
+            Metadata={HELIUM_METADATA: json.dumps(meta)}
+        )
+    except ClientError as e:
+        # suppress error from copying a file to itself
+        if str(e) == NO_OP_COPY_ERROR_MESSAGE:
+            return
+        raise
 
 def list_object_versions(path, recursive=True):
     bucket, key = split_path(path)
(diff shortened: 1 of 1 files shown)


api/python/t4/data_transfer.py
```diff
@@ -346,6 +347,7 @@ def delete_object(path):
         s3_client.head_object(Bucket=bucket, Key=key)  # Make sure it exists
         s3_client.delete_object(Bucket=bucket, Key=key)  # Actually delete it
 
+NO_OP_COPY_ERROR_MESSAGE = "An error occurred (InvalidRequest) when calling the CopyObject operation: This copy request is illegal because it is trying to copy an object to itself without changing the object's metadata, storage class, website redirect location or encryption attributes."
```
stevededalus: line length?

api/python/t4/data_transfer.py
```diff
@@ -359,13 +361,18 @@ def copy_object(src, dest, meta, version=None):
...
+    try:
+        s3_client.copy_object(
+            CopySource=src_params,
+            Bucket=dest_bucket,
+            Key=dest_key,
+            Metadata={HELIUM_METADATA: json.dumps(meta)}
+        )
+    except ClientError as e:
+        # suppress error from copying a file to itself
+        if str(e) == NO_OP_COPY_ERROR_MESSAGE:
```
stevededalus: Is this not an EntityAlreadyExists error? 

I'm confused about the specifics of when this arrises. Is this copying and s3 object to itself (same src and dest uri)?  Or is this copying an s3 object to an existing s3 object with identical underlying data (differing src and dest uri)?

Can you add some context to your PR description to make it clearer for reviewing?

api/python/t4/data_transfer.py
meffij: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length
specific exception for module-level constants

meffij: If you have an alternate convention you want me to follow, that'd be no problem. I just couldn't remember what the options are for splitting a long string

api/python/t4/data_transfer.py
meffij: it's a ClientException, because boto3 exceptions aren't the most helpful. The issue arises when you copy an object to itself without changing anything (metadata, storage class, etc.) I will document this in PR description

api/python/t4/data_transfer.py
meffij: I'm going to go with parens + implicit string joining as per the style guide

api/python/t4/data_transfer.py
stevededalus: It seems like we shouldn't make a call to copy an object onto itself, I'm trying to understand when it occurs and why we can't just always skip the s3 api call to copy.

I guess the flow is : pull down an existing manifest with remote s3 keys (not-materialized) and then modify the metadata and publish it back to s3. The copy can't know the metadata has been updated or not just from its current metadata. Is that the correct issue? If so, can we add that to the description.

api/python/t4/data_transfer.py
stevededalus: No, that;s fine. I'mm used to seeing this as string concated, but if its in the style guide and doesn't break pylint then it's cool.

stevededalus: That link says "Long string module level constants not containing whitespace that would be inconvenient to split across lines such as URLs or pathnames." There's a bunch of whitespace in that message

api/python/t4/data_transfer.py
stevededalus: It looks like a more stable way to do this is
```
['Error']['Code']
```
But is the error code is `InvalidRequest` then maybe it is too broad.

https://stackoverflow.com/questions/33068055/boto3-python-and-how-to-handle-errors

Add a todo for now to clean this up. Error messages are generally not stable across api calls and may have i18n issues.

meffij:
It seems like if no metadata is specified, then the metadata is copied from the source object. So we could no-op if `src` and `dest` are equal

PR #61, opened 2018-10-30
Use URLs in get, put, ls, delete

dimaryaz:
Note: this breaks the existing syntax! Try it and check if it's still ok.
In particular, I'm removing the `version` key from `get` cause it can now be part of the URL... but it's not very discoverable.

### DIFF
--- a/api/python/t4/api.py
+++ b/api/python/t4/api.py
@@ -13,7 +13,7 @@ from .data_transfer import (TargetType, copy_file, deserialize_obj, download_byt
                             upload_bytes, delete_object, list_objects,
                             list_object_versions, serialize_obj)
 from .packages import get_local_package_registry, get_package_registry
-from .util import (HeliumConfig, QuiltException, AWS_SEPARATOR, CONFIG_PATH,
+from .util import (HeliumConfig, QuiltException, CONFIG_PATH,
                    CONFIG_TEMPLATE, fix_url, parse_file_url, parse_s3_url, read_yaml, validate_url,
                    write_yaml, yaml_has_comments)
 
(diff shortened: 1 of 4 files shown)


api/python/t4/api.py
```diff
@@ -25,20 +25,34 @@ def copy(src, dest, meta=None):
...
 
 def put(obj, dest, meta=None):
-    if dest.endswith(AWS_SEPARATOR):
+    url = urlparse(dest)
+    if url.scheme != 's3':
+        raise NotImplementedError
+
+    bucket, path, version = parse_s3_url(url)
+    if version:
+        raise ValueError("Cannot push to a version")
```
kevinemoore: Should that error message say "put" instead of "push"?

api/python/t4/api.py
```diff
@@ -25,20 +25,34 @@ def copy(src, dest, meta=None):
...
+
+    if path.endswith(AWS_SEPARATOR):
```
kevinemoore: Minor thing, but it would be clearer to consistently use either '/' or AWS_SEPARATOR.

api/python/t4/api.py
```diff
@@ -82,22 +94,28 @@ def get(src, version=None):
...
-def delete(path):
-    """Delete an object from T4.
-
-    Does not delete local files.
+def delete(target):
+    """Delete an object.
 
     Parameters:
-        path (str): Path of object to delete
+        target (str): URI of the object to delete
```
stevededalus: This is somewhat opposed to what Aneesh wants, he wants to remove as much as possible "s3:/" from the API users, we should discuss.

PR #62, opened 2018-11-01
Local push

meffij:
TODO: unit test


api/python/t4/packages.py
```diff
@@ -527,6 +527,13 @@ def push(self, name, path):
             A new package that points to the copied objects
         """
         dest = fix_url(path)
+        if dest.startswith('file://'):
+            # local push
+            os.makedirs(path, exist_ok=True)
+            pkg = self._materialize(dest)
+            pkg.build(name)
```
stevededalus: probably want to start sanitizing `name` on push/build/etc. to prevent directory traversal now

api/python/t4/packages.py
```diff
@@ -527,6 +527,13 @@ def push(self, name, path):
...
+            return pkg
+
         if not dest.startswith('s3://'):
```
stevededalus: Probably should change the flow of these conditionals for readability. Possibly:

```
if 'file://':
...
elif 's3://' ''
...
else
raise NIE
```

api/python/t4/packages.py
stevededalus: so, my 1 concern here is that `build` and `push` implement the same logic wrt writing to the `package` and `named_packages` dirs, so it would be nice if this was a single function incase this logic needs to be updated/maintained (which it probably will for supporting arbitrary registries/dedupe etc).

I think the only reason it can't be just drop in replaced is that one version uses a tempfile and the other just uses a file handle locally. I think its ok to use a tempfile in both cases, although not ideal.

Refactoring that means you don't need to special case file and s3 in this method at all.

stevededalus: You `quote(name)` in init, so probably need to quote it here too

api/python/t4/packages.py
```diff
@@ -527,6 +527,13 @@ def push(self, name, path):
             A new package that points to the copied objects
         """
         dest = fix_url(path)
+        if dest.startswith('file://'):
+            # local push
+            os.makedirs(path, exist_ok=True)
+            pkg = self._materialize(dest)
```
stevededalus: when materializing to s3, we splat it into `{dest}/{pkg_name}` we probably should do the same for local disk too.

api/python/t4/packages.py
```diff
@@ -338,31 +338,50 @@ def get_meta(self, logical_key):
         entry = self._data[logical_key]
         return entry.meta
 
-    def build(self, name=None):
+    def build(self, name=None, registry=None):
         """
         Serializes this package to a local registry.
```
stevededalus: this is no longer local, I take it? update comment

api/python/t4/packages.py
```diff
@@ -527,48 +546,14 @@ def push(self, name, path):
...
         dest = fix_url(path)
-        if dest.startswith('file://'):
-            # local push
-            os.makedirs(path, exist_ok=True)
-            pkg = self._materialize(dest)
-            pkg.build(name)
+        if name:
+            dest = dest.strip('/') + '/' + quote(name) + '/'
+        if dest.startswith('file://') or dest.startswith('s3://'):
+            pkg = self._materialize(dest.strip('/'))
```
stevededalus: maybe `dest = fix_url(path).strip('/')` since it looks like you always strip it later

api/python/t4/data_transfer.py
```diff
@@ -490,6 +491,7 @@ def copy_file(src, dest, override_meta=None):
     if src_url.scheme == 'file':
         if dest_url.scheme == 'file':
             # TODO: metadata
+            os.makedirs(os.path.dirname(parse_file_url(dest_url)), exist_ok=True)
```
dimaryaz: I'd prefer something like `pathlib.Path(...).parent.makedirs()` - `pathlib` generally makes things simpler

PR #64, opened 2018-11-01
Formats Class

eode:
~Title pretty much says it.  Someone else handled the case where metadata is present -- this is just infers the `target` when you're accessing a blob that doesn't have one.~

Sadly, at first blush this now looks like I've just replaced three functions and an Enum with three base classes, a couple of subclasses, and quite a few instances.  This is because I have.  But, there are justifications for all that, and benefits to it.

One of the original goals for this was to allow users to add  their own serializers/deserializers.  But there are a lot of things we'll need to *really* make that work.

User deserializers for filenames can be pretty simple, but the reality is that if you don't make a formal interface for a particular area of code, and give that formal interface structural authority over what's added / etc, then things get kinda lax around the edges.  Edge cases get implemented multiple times, or something that's generally needful gets implemented only in one area.  

So, this:
* Provides a generic class that can be instantiated easily to add a new serializer/deserializer pair
* handles extensions automatically for deserialization, if provided at format creation time
* handled class types automatically for serialization, if provided at format creation time
* can be subclassed for special cases, like where lazy loading is preferred
  * does that with numpy/pandas/arrow
* provides filename extensions that *should* be used for some object that is being serialized, but is getting a new name
* updates metadata (if given at serialization time) for round-trip serialization
* can be ripped out and unceremoniously plopped into Quilt4, when that happens.
  * including whatever developed/contributed `Format` subclasses or additional `Format` configured instances that have been added in the mean time.
* can be expanded later to perform chunk-iterative serialization for formats that support it, if that ever becomes a request.

Notes:
This replaces `serialize_obj()`, `deserialize_obj()`, `_get_target_for_object()` and the `TargetTypes` enum.  There is now no *static* definition of target types, but also, serialization/deserialization is isolated from other code, and in general it shouldn't be necessary to access the target type anywhere but from the `formats` module.

ToDo:
* [x] Complain about a specific format if specified but not registered.
* [x] Match ext on api.put() if metadata not given.
* [x] Handle pandas serialization
* [x] Fix parquet serialization compatibility with select()


eode:
Tests need fixing..

eode:
I'm going to make this a bit more user-facing.

eode:
Could still use some test expansion, but 's decent.

eode:
This could use a look -- @dimaryaz ?

api/python/t4/formats.py
```diff
@@ -0,0 +1,415 @@
...
+    handled_types=[text_type],
+)
+
+
+class NumpyFormat(Format):
+    def __init__(self, name, handled_extensions=None, **kwargs):
+        handled_extensions = [] if handled_extensions is None else handled_extensions
+
+        # don't include these extensions unlss numpy is present.
+        if package_exists('numpy'):
```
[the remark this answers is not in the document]
eode: I have no problem with that.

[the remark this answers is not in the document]
eode: Oh -- but, it wouldn't fail, in that case -- or shouldn't. `package_exists` doesn't detect whether it's imported, it detects whether it's installed, but without spending the cost of importing it.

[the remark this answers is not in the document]
eode: Smart idea checking `sys.modules`, though.

api/python/t4/formats.py
```diff
@@ -0,0 +1,415 @@
+#! /usr/bin/python
```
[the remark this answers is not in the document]
eode: IDE Template.  I'll strike that.

api/python/t4/formats.py
```diff
@@ -0,0 +1,415 @@
...
+# -*- coding: utf-8 -*-
+
+""" formats.py
+
+This module handles binary formats, and conversion to/from objects.
+
+# Formats Class (plural)
```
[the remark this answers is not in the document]
eode: Good call.

api/python/t4/api.py
```diff
@@ -69,7 +70,7 @@ def get(src):
         tuple: ``(data, metadata)``.  Does not work on all objects.
     """
     data, meta = get_bytes(fix_url(src))
-    ext = pathlib.Path(src).suffix
+    ext = pathlib.PurePosixPath(src).suffix
```
dimaryaz: Here, you actually want to use the result of `fix_url(src)`, which will always use `/`. Also, `urlparse(...).path`. It can have a query string like `versionId=...`.

eode: Ah.  I thought we only accepted URLs as input, and didn't realize we were using query strings now.  S3 doesn't, from what I understand, and I didn't figure the filesystem would.  But if we're parsing them for our own use, that makes sense.  I probably should have accounted for the possibility anyways.

api/python/t4/api.py
```diff
@@ -52,6 +52,7 @@ def put(obj, dest, meta=None):
         meta (dict): Optional. metadata dict to store with ``obj`` at ``dest``
     """
     all_meta = {'user_meta': meta}
+    ext = pathlib.PurePosixPath(dest).suffix
```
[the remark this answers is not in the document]
eode: fixed.

api/python/t4/api.py
eode: fixed.

api/python/t4/formats.py
```diff
@@ -0,0 +1,418 @@
...
+    def match(cls, name):
+        """Match a format by exact name."""
+        return cls.registered_formats.get(name)
+
+    @classmethod
+    def for_ext(cls, ext, single=True):
+        """Match a format (or formats) by extension."""
+        ext = ext.lower().strip('. ')
+        if single:
+            matching_formats = cls.formats_by_ext[ext]
```
dimaryaz: Move it out of the `if`, since it's used in both cases

api/python/t4/formats.py
```diff
@@ -0,0 +1,418 @@
...
+        if single:
+            for format in reversed(cls.registered_formats.values()):
+                if format.handles_obj(obj):
+                    return format
+            return
+        return [format for format in reversed(cls.registered_formats.values()) if format.handles_obj(obj)]
+
+    @classmethod
+    def _get_format_name_from_meta(cls, meta):
+        name = None
```
dimaryaz: Shortcut:
```
name = meta.get('format', {}).get('name')
```

api/python/t4/formats.py
```diff
@@ -0,0 +1,418 @@
...
+            meta: object metadata, may contain deserialization prefs
+            **kwargs: metadata to (potentially) use.
+        """
+        # deactivated if this method is overridden.
+        if type(self).deserialize == Format.deserialize:
+            meta = {} if meta is None else meta
+            if getattr(self, '_deserializer', None) is None:
+                raise NotImplementedError()
+            format_meta = meta.get('format', {})
+            deserialization_kwargs = format_meta.get('deserialization', {})
```
dimaryaz: I'm not sure this is a good idea... This could potentially allow things like `np.load(..., allow_pickle=True)` (ok, looks like you're not using kwargs for numpy and parquet, but still.) But also, it's very Python-specific.

api/python/t4/formats.py
```diff
@@ -0,0 +1,418 @@
...
+
+NumpyFormat('numpy').register()
+
+
+class ParquetFormat(Format):
+    def __init__(self, name, handled_extensions=None, **kwargs):
+        handled_extensions = [] if handled_extensions is None else handled_extensions
+
+        if 'parquet' not in handled_extensions:
+            handled_extensions.append('parquet')
```
dimaryaz: Probably not good to modify caller's data... but really, I'd just hardcode the extensions for parquet and numpy for now.

api/python/tests/test_formats.py
```diff
@@ -0,0 +1,84 @@
+#! /usr/bin/python
```
dimaryaz: Remove this one

api/python/t4/formats.py
```diff
@@ -0,0 +1,418 @@
...
+        from pyarrow import parquet
+
+        buf = io.BytesIO(bytes_obj)
+        table = parquet.read_table(buf)
+        try:
+            obj = pa.Table.to_pandas(table)
+        except AssertionError:
+            # Try again to convert the table after removing
+            # the possibly buggy Pandas-specific metadata.
+            # XXX: Can we detect this during serialization and store it separately?
```
dimaryaz: Nope... it might not have been created by us. In any case, Pyarrow raising an assertion error on user input is a pyarrow bug - but they're not in a rush to fix it.

api/python/t4/api.py
```diff
@@ -51,13 +51,12 @@ def put(obj, dest, meta=None):
...
         meta (dict): Optional. metadata dict to store with ``obj`` at ``dest``
     """
-    data, target = serialize_obj(obj)
-    all_meta = dict(
-        target=target.value,
-        user_meta=meta
-    )
+    all_meta = {'user_meta': meta}
+    clean_dest = fix_url(dest)
+    ext = pathlib.PurePosixPath(urlparse(clean_dest).path).suffix
```
dimaryaz: Actually, sorry, one more ridiculous URL-related bug: need `unquote(urlparse(clean_dest).path)`.

(E.g., `s3://bucket/foo.%74x%74` is the same URL as `s3://bucket/foo.txt`)

api/python/t4/api.py
```diff
@@ -71,17 +70,11 @@ def get(src):
...
         tuple: ``(data, metadata)``.  Does not work on all objects.
     """
-    data, meta = get_bytes(fix_url(src))
-
-    target_str = meta.get('target')
-    if target_str is None:
-        raise QuiltException("No serialization metadata")
+    clean_src = fix_url(src)
+    data, meta = get_bytes(clean_src)
+    ext = pathlib.PurePosixPath(urlparse(clean_src).path).suffix
```
dimaryaz: Same here

api/python/t4/formats.py
eode: The former -- fine, but the too-python-specific part I can get on board with.  I'll change it.

eode: In consideration, even the former is rather an issue -- we shouldn't make it easy for mistakes to be made.

api/python/t4/formats.py
eode: Thanks for the reply.

api/python/t4/api.py
eode: Ugh.  yeah, URLs should technically be percent-decoded and then utf-8 decoded -- but then, shouldn't we only accept bytes for URLs, since an encoded URL is a byte-format type?

eode: ..but no, we're not taking bytes here, we're taking a unicode-encoded object -- and we should use the best format for internal manipulation, and expect that format from our users -- particularly since we're going to be used by hand in (for example) Jupyter notebooks, ipython, bpython, etc.

api/python/t4/api.py
eode: see comment above.

eode:
PR points addressed, re-merged with master (including integrating into `Bucket`).  I'm ready to merge this, pending approval of this one point:

The 'meta' dict is modified on serialization.  It could very reasonably be argued that info should instead be returned to the caller, and the calller should update that.  Instead, I took the stance that the `formats` module "owns" the `meta['format']` key.  This leads to shorter code when using it, but is more implicit.  The benefit is that it means less boilerplate on the caller's side.

..in any case, I'm really on the fence about that one point, and it's an easy change, so if anyone has thoughts on the matter, chime in.  Beyond that, this is done.

api/python/t4/api.py
stevededalus: I'm not following this thread. Isn't `unquote` not changing the encoding unless you pass it an optional argument? It's just remapping url escaped characters.  It looks like the the bucket api class we keep s3 urls unescaped internally and we should be consistent about that.

But looking at the bucket api, it looks like these are never escaped when accessing a bucket? Is the s3manager expecting unescaped strings @dimaryaz ?  If so, why ever escape/unescape?

api/python/tests/test_formats.py
```diff
@@ -0,0 +1,84 @@
...
+    metadata = [{} for o in objects]
+
+    for obj, meta in zip(objects, metadata):
+        assert FormatsRegistry.deserialize(FormatsRegistry.serialize(obj, meta), meta) == obj
+
+    df = pd.DataFrame([[1, 2], [3, 4]])
+    meta = {}
+    assert df.equals(FormatsRegistry.deserialize(FormatsRegistry.serialize(df, meta), meta))
+
+
```
stevededalus: extra white space here?

api/python/t4/api.py
eode: Unquote assumes that it's encoded as UTF-8 (which it's supposed to).  When quoting by spec, first the text is encoded as UTF-8, then encoded using percent encoding.  When unquoting, first the bytes are decoded using percent encoding, and then decoded via UTF-8.

The reason this is an issue is that we can't unquote more than once, ultimately.  We need to choose what our "external" or public-facing methods are, and what those accept.  I don't think this is a public-facing method.  If someone uses it, they should know what they're sending.

The only other consistent option is to do it right before needing to use the URL externally -- and that's problematic in our case because we do a lot of internal manipulation of the URLs.

I'd say that right now, we leave this as-is, and later convert everything over to an internal or external URI lib.  In the mean time, we *will* (and already do, in many areas) have inconsistencies with the URL spec.  `file://` when calling `Package.set()` is one example.

dimaryaz: You can try a path with a space and see what happens, e.g.:
```
urllib.parse.urlparse(fix_url('a b.txt')).path
```
You'll end up with `%20` - so need to `unquote` it.

api/python/t4/formats.py
```diff
@@ -0,0 +1,415 @@
...
+        """
+        if name is None:
+            raise TypeError("`name` is a required parameter.")
+        self.name = name
+        if serializer:
+            self._serializer = serializer
+        if deserializer:
+            self._deserializer = deserializer
+
+        if not (getattr(self, '_serializer', None) or (self.serialize != Format.serialize)):
```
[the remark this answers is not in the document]
dimaryaz: Can you refactor this? I'd really prefer an abstract base class and a subclass instead of attribute manipulation.

api/python/tests/test_formats.py
```diff
@@ -0,0 +1,78 @@
+#! /usr/bin/python
```
dimaryaz: Still not fixed?

eode: :-(  I removed that..  ..ah.  I thought your comment was in 'formats.py'.

eode: > So anyways... Until we switch to proper URI classes, can you add the unquote calls?

Definitely.

eode: > Can you refactor this?

..hokay.

eode: > Refactor

To be clear, we'll still probably want to do *something* to enable creation of simple formats by users.  That was made that way so that one call could create a format in cases where serialization / deserialization isn't a complex matter at all (JSON, Unicode, yaml, and similar formats that don't have a lot of underlying complexity and variance like CSV and some others do.  ..you have any complaints about a class factory function, then?

eode:
Added Pandas<-->CSV format. 
Fixed Parquet settings to be comaptible with s3 select (disabled whole-file compression, enabled per-column compression on all columns)

api/python/t4/formats.py
```diff
@@ -0,0 +1,505 @@
...
+                (and deserialized from) by this format
+        """
+        self.name = name if name else getattr(self, 'name', None)
+        if not self.name:
+            raise TypeError("No `name` attribute has been defined for {!r}".format(type(self).__name__))
+
+        # add user extensions if given
+        handled_extensions = [] if handled_extensions is None else handled_extensions
+        self.handled_extensions = getattr(self, 'handled_extensions', [])
+        self.handled_extensions.extend(ext for ext in handled_extensions if ext not in self.handled_extensions)
```
dimaryaz: This is O(N²)... but really, I'd rather get rid of the `getattr` above.

api/python/t4/formats.py
```diff
@@ -0,0 +1,505 @@
...
+
+        # add user types if given
+        handled_types = [] if handled_types is None else handled_types
+        self.handled_types = getattr(self, 'handled_types', [])
+        self.handled_types.extend(ext for ext in handled_types if ext not in self.handled_types)
```
dimaryaz: Same here

api/python/t4/formats.py
```diff
@@ -0,0 +1,505 @@
...
+    def handles_ext(self, ext):
+        """Check if this format handles the filetype indicated by an extension
+
+        Args:
+            ext: extension to check
+
+        Returns:
+            bool
+        """
+        exts = [e.lstrip().lower() for e in self.handled_extensions]
```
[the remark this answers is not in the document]
eode: Ok, I'll make `handled_extensions` private, too then.

api/python/t4/formats.py
eode: Sets are overkill, but sure.

api/python/t4/formats.py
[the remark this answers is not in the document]
eode: On second thought, I don't know.  We do want users to be able to see easily what extensions a format handles (if they have a format object for whatever reason).  ..and if they see that, they might add to it.  I could make it a property that returns a copy, but that seems like overkill.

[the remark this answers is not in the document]
eode: I suppose we'll just rely on users not modifying it (or modifying it properly) and leave it public.

api/python/t4/formats.py
eode: done

api/python/t4/formats.py
eode: done

api/python/t4/formats.py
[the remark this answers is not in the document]
eode: done

api/python/tests/test_formats.py
```diff
@@ -71,14 +74,74 @@ def test_formats_serdes():
...
+
+    # interpret first row as header, and first column as index
+    meta = {'format': {'name': 'csv', 'opts': {'use_index': True, 'use_header': True}}}
+    df1 = FormatRegistry.deserialize(test_data, meta)
+    bin, format_meta = FormatRegistry.serialize(df1, meta)
+    meta.update(format_meta)
+    df2 = FormatRegistry.deserialize(bin, meta)
+
+    assert test_data == bin
+    assert df1.equals(df2)
```
stevededalus: new line at end of file

api/python/t4/formats.py
```diff
@@ -0,0 +1,1001 @@
...
+        elif isinstance(value, str):
+            value = value.strip().lower()
+            map = {
+                'all': csv.QUOTE_ALL,
+                'minimal': csv.QUOTE_MINIMAL,
+                'none': csv.QUOTE_NONE,
+                'nonnumeric': csv.QUOTE_NONNUMERIC
+            }
+            return map.get(value, NOT_SET)
+        print("Unrecognized value for 'quoting' option: {!r}".format(value))
```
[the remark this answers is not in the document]
eode: Well, it's possible we'll get bad or outdated options in the future -- from someone messing with metadata, or from another client (if we do one in R, or similar).  I've changed it to a warning -- good?

..if there's a problematic or unknown option, this allows it to pass through and the serialization/deserialization to continue.

api/python/t4/formats.py
```diff
@@ -0,0 +1,1007 @@
...
+        handlers = cls.registered_handlers
+
+        # no duplicates, just reprioritize.
+        if handler in handlers:
+            handlers.pop(handlers.index(handler))
+
+        handlers.insert(0, handler)
+
+    @classmethod
+    def search(cls, obj_type=None, meta=None, ext=None, single=True, handlers=None):
```
dimaryaz: Please remove `single` from all the functions - it's weird to have a return type that depends on arguments.

If the list return type gets annoying, we can add helper functions that return a single item.

api/python/t4/formats.py
```diff
@@ -0,0 +1,1007 @@
...
+            elif name == 'na_values':
+                result_kwargs['na_rep'] = value[0]
+            else:
+                # exact match / pass through arg
+                result_kwargs[name] = value
+
+        return result_kwargs
+
+
+    def serialize(self, obj, meta=None, ext=None, raw_args=None, **format_opts):
```
dimaryaz: Remove `raw_args`; we can add it back if it comes up. It's safer to start with a minimal interface, and add things later.

api/python/t4/api.py
```diff
@@ -72,17 +72,11 @@ def get(src):
...
         tuple: ``(data, metadata)``.  Does not work on all objects.
     """
-    data, meta = get_bytes(fix_url(src))
-
-    target_str = meta.get('target')
-    if target_str is None:
-        raise QuiltException("No serialization metadata")
+    clean_src = fix_url(src)
+    data, meta = get_bytes(clean_src)
+    ext = pathlib.PurePosixPath(urlparse(unquote(clean_src)).path).suffix
```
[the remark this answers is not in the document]
eode: Good?

api/python/t4/formats.py
eode: Done.

api/python/t4/formats.py
eode: Done.

api/python/t4/packages.py
```diff
@@ -189,20 +195,21 @@ def deserialize(self):
...
-        except ValueError:
-            raise QuiltException("Unknown serialization target: %r" % target_str)
-
         physical_key = _to_singleton(self.physical_keys)
         data, _ = get_bytes(physical_key)
+
+        if func is not None:
+            return func(data)
+
+        pkey_ext = pathlib.PurePosixPath(urlparse(unquote(physical_key)).path).suffix
```
dimaryaz: Fix this unquote, too: `unquote(urlparse(physical_key).path)`

eode: Ugh.  I've been awful about those.

eode: Fixed, checked full changeset for any others, found one, fixed that as well.

PR #74, opened 2018-11-05
Change package methods to mutate existing package

meffij:


api/python/t4/packages.py
```diff
@@ -48,10 +47,10 @@ def get_package_registry(path=''):
     if path.startswith('s3://'):
         bucket = path[5:].partition('/')[0]
         return "s3://{}/.quilt".format(bucket)
-    else:
-        return get_local_package_registry().as_uri()
+    # default to local registry
```
stevededalus: Nit: Let's try to make comments close to sentences, capital first letter and punctuation at the end.

api/python/t4/packages.py
```diff
@@ -436,53 +434,49 @@ def set(self, logical_key, entry=None, meta=None):
...
             KeyError: when logical_key is not present to be deleted
         """
-        pkg = self._clone()
-        pkg._data.pop(logical_key)
-        # modified package should have new top hash
-        pkg._unset_tophash()
-        return pkg
+        self._data.pop(logical_key)
+        # must unset old top hash when modifying package
+        self._unset_tophash()
```
stevededalus: Can we add a test that the tophash changes when we do these sensitive operations (delete, set, update, capture)

PR #80, opened 2018-11-06
Fix coverage config

meffij:


stevededalus:
values looks the same in the bot though?

stevededalus:
Ah, I guess the bot runs against both instead of using an old cached version for master. That makes sense

PR #82, opened 2018-11-07
PackageEntry methods

meffij:


meffij:
I agree with Dima, making changes now

PR #83, opened 2018-11-07
Just remove _get() - nothing actually needs it

dimaryaz:


stevededalus:
lol, you need to test something else now, haha.

dimaryaz:
Whaaat. It's not like I removed coverage of existing code. It just doesn't like the overall coverage-to-code ratio decreasing?

PR #84, opened 2018-11-07
Change capture() to set_dir()

stevededalus:


api/python/t4/packages.py
```diff
@@ -326,35 +326,38 @@ def load(readable_file):
...
+    def set_dir(self, lkey, path):
         """
         Adds all files from path to the package.
 
         Recursively enumerates every file in path, and adds them to
             the package according to their relative location to path.
 
         Args:
-            path(string): path to package
+            lkey(string): prefix to add to every logical key.
```
dimaryaz: Maybe mention that an empty prefix is a valid input?

PR #87, opened 2018-11-07
Disallow empty URLs

dimaryaz:
Defaulting to "." is counter-intuitive. If you really want ".", then use ".".

### DIFF
--- a/api/python/t4/api.py
+++ b/api/python/t4/api.py
@@ -141,7 +141,7 @@ def list_packages(registry=None):
         A list of strings containing the names of the packages        
     """
     if not registry:
-        registry = ''
+        registry = 'whatever'
     registry = get_package_registry(fix_url(registry)).strip("/") + '/named_packages'
 
     registry_url = urlparse(registry)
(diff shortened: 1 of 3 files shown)


api/python/t4/api.py
```diff
@@ -141,7 +141,7 @@ def list_packages(registry=None):
         A list of strings containing the names of the packages        
     """
     if not registry:
-        registry = ''
+        registry = 'whatever'
```
dimaryaz: Obvious question: what _should_ this be? Probably not the current dir. `get_local_package_registry()`? We need to refactor some of that code.

stevededalus: why not the default local one?

api/python/t4/util.py
```diff
@@ -76,6 +76,9 @@ def split_path(path, require_subpath=False):
 
 def fix_url(url):
     """Convert non-URL paths to file:// URLs"""
+    if not url:
```
stevededalus: Don't we want an empty one to resolve to the directory at the current location

api/python/t4/api.py
```diff
@@ -141,7 +141,7 @@ def list_packages(registry=None):
         A list of strings containing the names of the packages        
     """
     if not registry:
-        registry = ''
```
stevededalus: What's the goal of this PR?

dimaryaz:
Fixing this in #95 instead

As for the goal: just sanity; empty URLs are an easy way to introduce bugs

PR #88, opened 2018-11-07
Basic Package.install

meffij:


api/python/t4/packages.py
```diff
@@ -195,6 +195,26 @@ def validate_package_name(name):
...
+            name(str): Name of package to install.
+            dest(str): Local path to download files to.
+                Defaults to $local_registry/files/{sha256 of file}
+            registry(str): Registry to install package to. Defaults to local registry.
+            pkg_hash(str): Hash of package to install. Defaults to latest.
+
+        Returns:
+            A new Package that points to files on your local machine.
+        """
+        pkg = Package(name=name, pkg_hash=pkg_hash, registry=registry or '')
```
stevededalus: `Package.browse` now instead of `Package`. Otherwise looks good

PR #89, opened 2018-11-07
Add registry argument to push

meffij:


stevededalus:
lgtm, but waiting for Dima's change that lets the build with arbitrary registry work

PR #95, opened 2018-11-07
Implement registries at arbitrary locations

dimaryaz:
Also hide `get_package_registry` from the user: the user should pass in normal URLs or None, and we'll append '.quilt' or get the default registry as needed.

One weirdness that came up: the `registry` arg in `build` is just a normal path. So we two methods like this:
- `build(name, path)`
- `push(path, name)`
Should probably pick one convention.

### DIFF
--- a/api/python/t4/__init__.py
+++ b/api/python/t4/__init__.py
@@ -14,4 +14,4 @@ from .api import (
     config,
 )
 
-from .packages import Package, get_local_package_registry, get_package_registry
+from .packages import Package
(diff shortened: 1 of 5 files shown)


dimaryaz:
I still can't quite make sense of how `push` and `install` are supposed to work...

E.g., `install` operates on two registries: it looks up a package, then pushes it. And I'm guessing in the common case, those two registries _will_ be different: look up in S3, push to local. How does the user specify all of that?

ResidentMario:
So we talked this out, and I think that the `dest` and `dest_path` arguments are purely engineering limitations that we can live with for now.

But we'll want to get rid of these bonus arguments ASAP once we land this current API revision.

PR #97, opened 2018-11-07
Don't keep top hash in memory

meffij:


api/python/t4/packages.py
```diff
@@ -619,16 +620,14 @@ def delete(self, logical_key):
             KeyError: when logical_key is not present to be deleted
         """
         self._data.pop(logical_key)
-        # Must unset old top hash when modifying package
-        self._unset_tophash()
         return self
 
     def _top_hash(self):
```
stevededalus: Any reason to keep the private and public version instead of just having it all be the public one?

meffij: probably not; let me check

PR #98, opened 2018-11-07
Make brackets throw when resulting in an empty package

meffij:


api/python/t4/packages.py
```diff
@@ -338,6 +338,10 @@ def __getitem__(self, prefix):
             if key.startswith(slash_prefix):
                 new_key = key[len(slash_prefix):]
                 result.set(new_key, entry)
+
+        if not result._data:
+            raise KeyError("Slice not found.")
```
stevededalus: s/Slice/Package Slice

PR #99, opened 2018-11-07
Package.fetch()

akarve:
* refactor some repeated, checks for singleton physical keys
* move ~`Package.copy()`~ to `Package._copy()` (private)


api/python/t4/packages.py
```diff
@@ -340,6 +346,25 @@ def __getitem__(self, prefix):
...
+
+        Args:
+            dest: where to put the files (locally)
+
+        Returns:
+            None
+        """
+        # TODO: do this with improved parallelism? connections etc. could be reused
+        for key, entry in self._data.items():
+            entry.fetch(os.path.join(dest, key))
```
dimaryaz: `dest` should probably be treated as a URL, not a path? That's how it works in `PackageEntry.fetch`

PR #100, opened 2018-11-07
Add a PackageEntry level set for overwriting physical keys

stevededalus:
intended for use for calling on the result of a [] call

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -102,7 +102,6 @@ class PackageEntry(object):
         Sets the user_meta for this PackageEntry.
         """
         self.meta['user_meta'] = meta
-        # return self
 
     def get_user_meta(self):
         """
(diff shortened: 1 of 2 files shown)


api/python/t4/packages.py
```diff
@@ -120,6 +111,7 @@ def set_user_meta(self, meta):
         Sets the user_meta for this PackageEntry.
         """
         self.meta['user_meta'] = meta
+        # return self
```
dimaryaz: Remove it?

api/python/t4/packages.py
```diff
@@ -143,6 +135,46 @@ def _verify_hash(self, read_bytes):
...
+                else in the entry
+
+        Returns:
+            self
+        """
+        if path is None and meta is None:
+            raise PackageException('Must specify either path or meta')
+
+        if path is None:
+            return self.set_user_meta(meta)
```
dimaryaz: Looks weird cause `set_user_meta` doesn't return anything.

I'd do something like this, to make it more uniform:
```
if path is not None:
  ...
if meta is not None:
  ...
```

Also, now that I think about, it's kinda weird to have:
- `set`
- `set_user_meta`
- `_set_path`
Maybe it should be `set_path`. Or maybe we should get rid of `set_user_meta` cause `set` is sufficient.

stevededalus: Restructured it to follow your advice, but leaving method signatures for now. I think `set` and `set_user_meta` were minimum API requirements.

PR #102, opened 2018-11-08
threaded upload for _materialize(), and therefore push().

eode:
This was pretty simple:

* `concurrent.futures`
* `ThreadPoolExecutor`, because a Process Pool would be way overkill, and it's network, not CPU bound
* TQDM Just Magically Worked -- multiline display and all.


eode:
Well, nested `ThreadPoolExecutor`s isn't really an issue, so long as it's not for something that leads to exponentially increasing thread count.  The basic situation is:

* Is it worth it to parallelize `push()`? ..it seems so.
* For any given push, is it worth it to parallelize etag iteration and object version listing? ..it seems so.

..the result should, then, also be worth it -- to an extent.  Since we're basically focusing on overcoming io-bound limits, we don't need a ton of workers on the task, so long as for any given IO source, there's at least one item running.

Potential Solutions:
* Limit both `ThreadPoolExecutor`s `max_workers`
* Limit `upload_file()` to max_workers=2, since it's primarily parallelizing reading files and network IO anyways
* Pass a `_max_workers` param on down the line, from `pkg._materialize()` to `pkg._copy()` to `data_transfer.copy_file()` to  `data_transfer.upload_file()`
  * 🤨
* Keep some kind of global state
  * 😾
* Don't parallelize `Package.push()`

Recommendation:
* Limit both `ThreadPoolExecutor`s
  * `upload_file()` to 1 or 2 additional threads, as it is primarily serializing two IO tasks, not CPU ones
    * though hashing can be CPU intensive, threading doesn't help, as it's all on the same CPU core.
  * `push` to 2 or 3, as it is primarily just trying to keep IO going while other network connections are being established

@dimaryaz 
Since I'm poking around at other peoples' code, can I get an approval on that thought?  

thanks..

dimaryaz:
Looks like it's not quite as bad - the thread count won't be an issue cause in the end, we're using a global S3 transfer manager that has a single thread pool. But, we'd still end up with N progress bars when uploading N files, which isn't a great UI.

In any case, I'd still like a cleaner solution. The `upload_file` and `download_file` already have all the logic to upload/download multiple files in parallel - however, they only do that for directory uploads and downloads. We could refactor them to accept a list of files (or a list of tuples of files with whatever extra data), and then make directory uploads/downloads use those functions.

dimaryaz:
So to clarify: the `upload_file`, `download_file`, `copy_file`, etc. in `data_transfer.py` are not public APIs, so it's fine to change them however you want.

JacksonMaxfield:
Wanted to jump in just out of curiosity, I haven't used `concurrent` before but what benefit does it bring over `multiprocessing.dummy.Pool`?

Also generally, in the past on my work I have used `os.cpu_count()` to get "number of threads that the system can safely handle". Generally this is _okay_. I noticed that my personal machine has only 2 cpus but I bump my thread limit up to 30 in most cases with a parameter, but for across the board `os.cpu_count()` works generally alright, as long as you have a cap put on it. In my case, I had to cap it because some of the machines in AICS have upwards of 120 cores and some of our internal services begin having io bottleneck issues with that many threads open to the service.

eode:
> Wanted to jump in just out of curiosity, I haven't used `concurrent` before but what benefit does it bring over `multiprocessing.dummy.Pool`?

Nothing, from a technical perspective, but:
* It's overall a more united API
* it's more canonical, and has better documentation
* it uses futures, which are a pretty convenient interface

> Also generally, in the past on my work I have used `os.cpu_count()` to get "number of threads that the system can safely handle". Generally this is _okay_. I noticed that my personal machine has only 2 cpus but I bump my thread limit up to 30 in most cases with a parameter, but for across the board `os.cpu_count()` works generally alright, as long as you have a cap put on it. In my case, I had to cap it because some of the machines in AICS have upwards of 120 cores and some of our internal services begin having io bottleneck issues with that many threads open to the service.

Well, in this case, we have (comparatively fast) code, and (comparatively slow) IO activity.  So, as long as our IO activity is continuous, we're good, and two to three threads achieve that.  Beyond that, your gains (if any) are typically from stressing the io systems -- stealing disk bandwidth from other procs, and network bandwidth from other local users, which is generally a no-no.  Of course, that's something that can be made possible, but it's something we shouldn't default to.

akarve:
Now implemented in #204

PR #105, opened 2018-11-08
Use / for package root in set_dir

meffij:


meffij:
Reopening to prevent bug that leads to leading `/` in logical keys

PR #107, opened 2018-11-09
First draft of diff

meffij:


api/python/t4/packages.py
```diff
@@ -83,6 +83,14 @@ def __init__(self, physical_keys, size, hash_obj, meta):
         self.hash = hash_obj
         self.meta = meta
 
+    def __eq__(self, other):
+        return (
+            self.physical_keys == other.physical_keys
```
stevededalus: physical keys too?

api/python/t4/packages.py
```diff
@@ -721,3 +729,34 @@ def _materialize(self, path):
...
+            added, modified, deleted (all lists of logical keys)
+        """
+        added = []
+        modified = []
+        self_keys = self._data.keys()
+        other_keys = other_pkg._data.keys()
+        for lk in self_keys:
+            if lk not in other_keys:
+                added.append(lk)
+            elif not self._data[lk] == other_pkg._data[lk]:
```
stevededalus: probably don't want to diff the phyiscal keys in the package entries?

api/python/t4/packages.py
```diff
@@ -721,3 +729,34 @@ def _materialize(self, path):
             # Treat as a local path
             pkg.set(logical_key, new_entry)
         return pkg
+
+    def diff(self, other_pkg):
+        """
+        Returns three lists -- added, modified, deleted.
+
+        Added: present in other_pkg but not in self.
```
stevededalus: Looks like you're logic is the reverse of these case, added is in self_keys but not other and deleted is in other but not self.

PR #108, opened 2018-11-09
Throw when set_dir doesn't add anything

meffij:


api/python/t4/packages.py
```diff
@@ -443,16 +443,20 @@ def set_dir(self, lkey, path):
         if url.scheme == 'file':
             src_path = pathlib.Path(parse_file_url(url))
             files = src_path.rglob('*')
+            has_added_file = False
```
stevededalus: I'm not sure this logic is correct, I think that `set_dirs` should except empty directories but raise exceptions on directories that don't exist?

akarve:
It should be legal to call `.set_dir(LKEY, EMPTY_DIR)`. And the reason is, the author may not know at packaging time if the directory is empty. E.g. `proj/errors` might be empty sometimes and not others. So an empty `proj/errors` is meaningful and we should support it.

~`set_dir(LKEY, NON_EXISTENT_DIR)`~ should fail b/c there's literall nothing there.

meffij:
Consensus is to check if DIR exists and is a directory, and throw if it doesn't or isn't

PR #113, opened 2018-11-14
Bucket API

meffij:


ResidentMario:
I and .@akarve looked at this implementation sketch earlier today and it looked 👍.

ResidentMario:
Breaking my feedback up section by section. cc @akarve.

### 1. bucket dictionary indexing
The proposed `Bucket` implementation doesn't allow navigating into a `Bucket` using dictionary key access, as is allowed on packages. This has implications on all of the API methods. For example, to `deserialize` an object, you have to pass the path to the object into the method.

Long-term we want `Bucket` and `Package` to inherit from the same base class anyway, so we'll need to work out how to make this possible. However I think this is a separate set of PRs, since we'll need to tweak `Package` as well to accommodate this (I have some notes on what I mean by this that'll go in a separate PR).

### 2. `keys`
Right now keys returns a full scan of the entire bucket as a list, essentially running `stat` on the whole store. This will have terrible performance for large buckets. We could potentially get around this by returning an iterator. Backlogging investigating an iterator for later.

### 3. `set` versus `put`

I propose that in general, if there is a `set` method on `Package`, it makes sense to call that method `put` on `Bucket`. Specifically:

* `set` <-> `put`
* `set_dir` <-> `put_dir`
* `set_meta` <-> `put_meta`

Proposed changes:
* Rename `Bucket.put_file` to `Bucket.put`.
* Rename `Bucket.put` to `Bucket.serialize` (to match `Package.deserialize`).
* Rename `Bucket.set_meta` to `Bucket.put_meta`.

akarve:
@ResidentMario Can we leave the API names as is and backlog any naming changes? I think the "API Coherence" design is correct enough for now.

akarve:
@ResidentMario Bucket dictionary indexing can go in backlog and might be easy with Dima's new changes

ResidentMario:
Bucket dictionary indexing is probably blocked by implementing a base class, which is probably out of scope for the `0.4.0` release.

akarve:
Regarding `keys` afaik `list_objects` is limited to 1K results and pages?

api/python/t4/bucket.py
```diff
@@ -0,0 +1,221 @@
...
+        """
+        Creates a Bucket object.
+
+        Args:
+            bucket_uri(str): URI of bucket to target. Must start with 's3://'
+
+        Returns:
+            a new Bucket
+        """
+        if not bucket_uri.startswith('s3://'):
```
dimaryaz: Just use `parse_s3_url`; doing string manipulation is very fragile

api/python/t4/bucket.py
```diff
@@ -0,0 +1,221 @@
...
+            meta(dict): optional user-provided metadata to store
+        """
+        dest = self._uri + key
+        meta = meta or {}
+        data, target = serialize_obj(obj)
+        all_meta = dict(
+            target=target.value,
+            user_meta=meta
+        )
+        put_bytes(data, fix_url(dest), all_meta)
```
dimaryaz: Don't need `fix_url` - you already know it's well-formed

api/python/t4/bucket.py
```diff
@@ -0,0 +1,221 @@
...
+        if not src_path.is_dir():
+            raise QuiltException("Provided directory does not exist")
+
+        files = src_path.rglob('*')
+        for f in files:
+            if not f.is_file():
+                continue
+            new_key = key + f.relative_to(src_path).as_posix()
+            new_path = self._uri + new_key
+            copy_file(f.resolve().as_uri(), new_path)
```
dimaryaz: `copy_file` already supports directories

api/python/t4/bucket.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def keys(self):
+        """
+        Lists all keys in the bucket.
+
+        Returns:
+            list of strings
+        """
+        return list(map(lambda x: x.get('Key'), list_objects(self._bucket, '')))
```
dimaryaz: `[x.get('Key') for x in list_objects(...)]`

api/python/t4/bucket.py
```diff
@@ -0,0 +1,221 @@
...
+
+        Raises:
+            if path doesn't exist
+            if download fails
+        """
+        if not key.endswith('/'):
+            copy_file(self._uri + key, fix_url(path))
+            return
+
+        _download_dir(self._bucket, key, fix_url(path))
```
dimaryaz: Just use `copy_file` for everything - it already checks for `/` and downloads dirs.

You can also use `download_file` to avoid dealing with URLs.

api/python/t4/bucket.py
```diff
@@ -0,0 +1,221 @@
...
+        Raises:
+            if download fails
+        """
+        response = s3_client.head_object(
+            Bucket=self._bucket,
+            Key=key
+        )
+        return response['Metadata']
+
+    def set_meta(self, key, meta):
```
dimaryaz: `set_user_meta`?

api/python/t4/bucket.py
```diff
@@ -0,0 +1,221 @@
...
+        new_meta['user_meta'] = json.dumps(meta)
+        s3_client.copy_object(
+            CopySource={
+                "Bucket": self._bucket,
+                "Key": key
+            },
+            Bucket=self._bucket,
+            Key=key,
+            MetadataDirective='REPLACE',
+            Metadata=new_meta
```
dimaryaz: Pretty sure this doesn't work - need the "helium" header. But really, just use the existing `copy_object` helper function.

api/python/t4/bucket.py
```diff
@@ -0,0 +1,221 @@
...
+        Args:
+            key(str): key in bucket to get meta for
+
+        Returns:
+            dict of meta
+
+        Raises:
+            if download fails
+        """
+        response = s3_client.head_object(
```
dimaryaz: I'd like to keep all low-level S3 code in `data_transfer.py` - so add a helper function for getting the metadata.

Also, you'll need to parse the correct header. See `_parse_metadata`.

api/python/tests/test_bucket.py
```diff
@@ -0,0 +1,56 @@
+import botocore.session
+from botocore.stub import Stubber, ANY
```
akarve: `ANY` isn't used? also: run pylint

akarve: I think it makes sense to add something like `test_bucket_io` which tests, e.g. a `put`/`fetch` pair. I realize that `Stubber` isn't _that_ useful because we have to specify all of the responses but it seems strange not to test any of the IO methods.

api/python/t4/bucket.py
ResidentMario: `t4.Package` has no `set_meta` but it does have a `get_meta` so it makes more sense to me to call this `set_meta` for now (but see my other comment about a broader rename).

api/python/t4/bucket.py
```diff
@@ -0,0 +1,197 @@
...
+            None
+
+        Raises:
+            if no file exists at path
+            if copy fails
+        """
+        dest = self._uri + key
+        copy_file(path, dest)
+
+    def put_dir(self, key, directory):
```
dimaryaz: Still not convinced that it's needed, given that `put_file` also handles directories... but no strong objections.

api/python/t4/data_transfer.py
```diff
@@ -531,3 +531,27 @@ def get_bytes(src):
...
+        'Bucket': bucket,
+        'Key': key
+    }
+    if version_id:
+        params['VersionId'] = version_id
+    resp = s3_client.head_object(**params)
+    meta = _parse_metadata(resp)
+    return meta
+
+def set_meta(dest, meta):
```
dimaryaz: This one doesn't touch low-level S3 stuff, and just calls other helpers - so I'd rather delete it and move the code into `Bucket.set_meta`.

(Also, nothing else in `data_transfer.py` even knows about "user_meta", so doesn't seem like the right level of abstraction here.)

PR #117, opened 2018-11-15
Convert packages from flat lists to trees

dimaryaz:


ResidentMario:
> dimaryaz wants to merge 1 commit into quote from package_tree

Presumably this should be to `master`?

dimaryaz:
I just wanted to merge #115 first, then rebase this PR onto master.

(The fix in #115 is unrelated to this one, so I sent it out as a separate PR - but I also didn't want to deal with merge conflicts, so made this change on top of the other one. Don't know if there's a better way?)

api/python/t4/packages.py
```diff
@@ -238,9 +238,9 @@ def __call__(self):
 class Package(object):
     """ In-memory representation of a package """
 
-    def __init__(self, data=None, meta=None):
-        self._data = {} if data is None else data
-        self._meta = {'version': 'v0'} if meta is None else meta
+    def __init__(self, children=None, meta=None):
+        self._children = children or {}
+        self._meta = meta or {'version': 'v0'}
```
ResidentMario: Doesn't this mean that if the `meta` parameter is non-null, no `version` tag will be populated?

api/python/t4/packages.py
```diff
@@ -378,20 +373,27 @@ def fetch(self, dest):
...
 
     def __iter__(self):
-        return iter(self._data)
+        return iter(self._children)
 
     def __len__(self):
-        return len(self._data)
+        return len(self._children)
+
+    def _walk(self, prefix=''):
```
ResidentMario: If we expose this as a public API method, does this also solve "[Implement a walk method on Package that traverses the package as a tree](https://app.asana.com/0/855836770921302/911647053675713)"?

api/python/t4/packages.py
dimaryaz: Yeah... but I didn't change that - just made the code simpler. I'll undo it, though, just in case.

api/python/t4/packages.py
dimaryaz: Asana is down for me, so can't see the task. What do we want the walk method to look like?

api/python/t4/packages.py
```diff
@@ -316,12 +316,15 @@ def _from_path(cls, uri):
...
-
-    def _clone(self):
-        """
-        Returns clone of this package.
-        """
-        return self.__class__(copy.deepcopy(self._data), copy.deepcopy(self._meta))
+    @classmethod
+    def _split_key(cls, logical_key):
+        if isinstance(logical_key, string_types):
+            path = logical_key.split('/')
```
akarve: minor: we shouldn't we have a global `QUILT_SEPARATOR` or something. maybe proper URI handling will fix this.

api/python/t4/packages.py
```diff
@@ -618,30 +597,29 @@ def set(self, logical_key, entry=None, meta=None):
...
-            if meta is not None:
-                entry.meta = meta
-            self._data[logical_key] = entry
         elif isinstance(entry, PackageEntry):
-            if meta is not None:
-                raise PackageException("Must specify metadata in the entry")
-            self._data[logical_key] = entry
+            entry = entry._clone()
         else:
             raise NotImplementedError("Needs to be of type str")
```
akarve: Missing a word for clarity `"Expected a string for X"` or `"X needs to be  of type str"`

api/python/t4/packages.py
```diff
@@ -668,9 +648,10 @@ def top_hash(self):
         top_hash = hashlib.sha256()
         hashable_meta = copy.deepcopy(self._meta)
         hashable_meta.pop('top_hash', None)
+        print('%r' % hashable_meta)
```
akarve: do we want this `print`?

api/python/t4/packages.py
ResidentMario: (Asana is back up)

I tested this out in a notebook, and think this is exactly the intended `t4.Package.walk` behavior: a function that provides an iterator over the `(path, entry)` tuples.

So yeah, I suggest we expose this as a public method.

api/python/t4/packages.py
```diff
@@ -617,30 +596,29 @@ def set(self, logical_key, entry=None, meta=None):
...
+        path = self._split_key(logical_key)
+
+        pkg = self._ensure_subpackage(path[:-1])
+        pkg._children[path[-1]] = entry
 
-    def _update_meta(self, logical_key, meta):
-        self._data[logical_key].meta = meta
         return self
 
+    def _ensure_subpackage(self, path):
```
meffij: Please write a docstring that specifies at least the types of the arguments and the type this returns

api/python/t4/packages.py
meffij: Please write a docstring with at least types in and out

api/python/t4/packages.py
```diff
@@ -316,12 +316,15 @@ def _from_path(cls, uri):
...
         return pkg
 
-
-    def _clone(self):
-        """
-        Returns clone of this package.
-        """
-        return self.__class__(copy.deepcopy(self._data), copy.deepcopy(self._meta))
+    @classmethod
+    def _split_key(cls, logical_key):
```
meffij: Please write a docstring with at least types in and out

api/python/t4/packages.py
```diff
@@ -722,17 +702,17 @@ def _materialize(self, path):
             fail to put bytes
             fail to put package to registry
         """
-        pkg = self._clone()
+        pkg = self.__class__({}, self._meta)
```
meffij: Why is this better than having a `._clone` method?

dimaryaz:
@eode: wow, it never occurred to me to try that. Anyways, the fact that `dict(Package(...))` works is actually an accident; I'd prefer it didn't, but not sure if that's possible. The args to the `Package` constructor weren't meant to be a public API... but they're not really used, so I removed them to avoid any confusion. Eventually, we probably want to support `Package(Package(...))`, etc.

PR #121, opened 2018-11-15
Make `PackageEntry.deserialize()` take deserializer override

eode:
Allow an argument for deserialize, so that:
`des(p.get(k))`
can be written as:
`p[k].deserialize(des)`
..and therefore, the nice, natural, and succinct:
`p[k](des)`

* The provide deserializer overrides any deserializer specified in the metadata.
* PackageEntry deserialization failures now reference `PackageEntry.deserialize` docs.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -187,28 +187,45 @@ class PackageEntry(object):
         """
         return _to_singleton(self.physical_keys)
 
-    def deserialize(self):
+    def deserialize(self, deserializer=None):
         """
         Returns the object this entry corresponds to.
+
+        Custom Deserializers:
+            If `deserializer` is given, it is passed the physical key
+            of the data that would have been deserialized, then its
+            return value is returned.
+
+            Example:
+                p[k](lambda k: json.loads(Path(urlparse(k).path).read_text()))
+
+        Args:
+            deserializer: use the given function for deserialization.
 
         Returns:
             The deserialized object from the logical_key
 
         Raises:
-            physical key failure
-            hash verification fail
-            when deserialization metadata is not present
+            KeyError: physical key failure
+            QuiltException: Unknown deserializer
+            QuiltException: hash validation failure
+            QuiltException: when deserialization metadata is not present
         """
+        physical_key = _to_singleton(self.physical_keys)
+        if deserializer:
+            return deserializer(physical_key)
+
         target_str = self.meta.get('target')
         if target_str is None:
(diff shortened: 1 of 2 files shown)


api/python/t4/packages.py
```diff
@@ -187,28 +187,49 @@ def get(self):
         """
         return _to_singleton(self.physical_keys)
 
-    def deserialize(self):
+    def deserialize(self, deserializer=None):
         """
         Returns the object this entry corresponds to.
 
+        Custom Deserializers:
+            If `deserializer` is given, it is passed the physical key
```
dimaryaz: Just wondering if passing bytes, rather than the physical key, would be more useful? Of course, reading a file directly can be more efficient than reading it into memory, then parsing - but this requires the user to parse our file:// and s3:// URLs and figure out what to do with them.

Maybe a good solution is to expose `get_bytes`, etc. as an API - `t4.get_bytes` or something?

api/python/t4/packages.py
```diff
@@ -187,28 +187,49 @@ def get(self):
...
+            of the data that would have been deserialized, then its
+            return value is returned.
+
+            Example for decoding a json blob:
+                >>> des = lambda k: json.loads(Path(urlparse(k).path).read_text())
```
dimaryaz: This is already pretty complicated - but still not enough. E.g., need to `unquote` the `.path`. There's a helper function for this, `parse_file_url` - though it's not really documented as a user API.

api/python/t4/packages.py
```diff
@@ -187,28 +187,49 @@ def get(self):
...
+                >>> obj = p[k](des)  # obj is the deserialized JSON blob
```
dimaryaz: I know I brought this up in Quilt2... but this is the same thing as:
```
obj = des(pk[k].get())
```
In fact, at that point, no need to create a lambda - just call:
```
obj = json.loads(...(p[k].get())
```
So I'm still not convinced that passing a deserializer is actually useful.

Though I could imagine it being useful if e.g. we were passing an open file descriptor into `des`: then we'd handle opening/closing the file, and the user wouldn't have to worry about that.

meffij:
I agree with Dima that this doesn't seem to accomplish much for our users as written.

akarve:
I will investigate and report back. No firm opinion yet but the albeit redundant syntax of Quilt2 `.des()` was popular with some key users.

eode:
@akarve noted, will do.

PR #122, opened 2018-11-15
Add commit on push and build

ResidentMario:
`push` and `build` now accept an optional `message` `kwarg`.

On `push` and `build`, this `kwarg` will be inserted into the package metadata as a `commit_message: message` key-value pair. If no message is specified, `message` will be `None`.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -519,6 +519,24 @@ class Package(object):
         entry = self._data[logical_key]
         return entry.meta
 
+    def _set_commit_message(self, msg):
+        """
+        Sets a commit message.
+
+        Args:
+            msg: a message string
+
+        Returns:
+            None
+
+        Raises:
+            msg is not a string
+        """
+        if not isinstance(msg, str):
+            raise ValueError("The commit message must be a string.")
+
+        self._meta.update({'commit_message': msg})
+
     def build(self, name=None, registry=None, message=None):
         """
         Serializes this package to a registry.
@@ -532,7 +550,7 @@ class Package(object):
         Returns:
             the top hash as a string
         """
-        self._meta.update({'commit_message': message})
+        self._set_commit_message(message)
 
         registry_prefix = get_package_registry(fix_url(registry) if registry else None)
 
@@ -697,8 +715,7 @@ class Package(object):
             A new package that points to the copied objects
         """
         self.validate_package_name(name)
-
-        self._meta.update({'commit_message': message})
+        self._set_commit_message(message)
 
         if registry is None:
             registry = dest
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -539,10 +539,10 @@ class Package(object):
             None
 
         Raises:
-            msg is not a string
+            a ValueError if msg is not a string
         """
         if msg is not None and not isinstance(msg, str):
-            raise ValueError("The commit message must be a string.")
+            raise ValueError("The package message must be a string.")
 
         self._meta.update({'message': msg})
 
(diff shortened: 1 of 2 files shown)


akarve:
can you merge latest master? it should fix CI.
also, unit test recommended.

ResidentMario:
Merge done. I can add a unit test, it's a bit of an internal do-nothing field for the moment though.

akarve:
generally, all PRs have unit tests unless there is a specific reason (like a docs change)

akarve:
I think this is a bug:
```
with open("./myman.json", "w") as file:
    p.dump(file)
```
`myman.json` shows `"commit_message": null`

also: can we just call it `message:`?

ResidentMario:
Renamed to `message`, but I can't reproduce this problem:

```
t4.Package().set("foo", "foo").build("foo/bar", message="foo")
with open("test.json", "w") as f:
    t4.Package.browse("foo/bar").dump(f)
!cat test.json

{"version": "v0", "message": "foo", "top_hash": {"alg": "v0", "value": "47d1c5b88296162a7fadb63ed31fe49ab1a994f0e97161772e4b73b6049f2933"}}
{"logical_key": "foo", "physical_keys": ["file:///Users/alex/Desktop/temp/foo"], "size": 0, "hash": {"type": "SHA256", "value": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"}, "meta": {}}
```

```
p = t4.Package().set("foo", "foo")
p.build("foo/bar", message="foo")
with open("test.json", "w") as f:
    p.dump(f)
!cat test.json

{"version": "v0", "message": "foo", "top_hash": {"alg": "v0", "value": "47d1c5b88296162a7fadb63ed31fe49ab1a994f0e97161772e4b73b6049f2933"}}
{"logical_key": "foo", "physical_keys": ["file:///Users/alex/Desktop/temp/foo"], "size": 0, "hash": {"type": "SHA256", "value": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"}, "meta": {}}
```

Both of these have a `message` populated.

akarve:
I think we will want to merge this after #131; and revise it to store message in the dir-level metadata for `/`, package root?

ResidentMario:
Well I was originally hoping to get this PR merged in first, but with the current state of affairs that's the correct course of action yeah.

meffij:
@akarve @ResidentMario Top-level metadata is actually not impacted by dir-level metadata at all

ResidentMario:
@meffij Looks like it does. Because the message goes to the top-level metadata it seems to mutate a component of the tophash, which is causing a test to fail because the tophash you get before you build (without a commit message) and the tophash you get after (with a tophash) is different.

meffij:
That sounds correct, and that wouldn't be impacted by directory-level metadata. I was trying to say this PR does not depend on 131, because none of the stuff 131 does affects top-level metadata.

stevededalus:
> I think we will want to merge this after #131; and revise it to store message in the dir-level metadata for `/`, package root?

Why have it on package root instead of top level?

stevededalus:
This is a little strange, you're writing this as a commit message, but we have no notion of package history so it's really just a package description. The commit messages don't get append, the most recent wipes the old ones out. So if I modify two files in a set of 1,000 and commit with "Fix xyz.jpeg" that produces a package that just has message="Fix xyz.jpeg", that doesn't really mean anything.

ResidentMario:
We discussed offline renaming the field to `description`. The point that there aren't truly meaningful versions is valid. After thinking about it for a time, I decided to keep it at `message` though, because `description` is an awkward argument for a `push` or `build` command. We can reinforce that it's really a per-version thing in our best practices.

PR #124, opened 2018-11-16
Rename helium to t4 and pyarrow to parquet

dimaryaz:
- Rename the metadata header and xattr from "helium" to "t4"
- Rename "PYARROW" to "PARQUET", since Parquet is the name of the format

NOTE: This breaks metadata on all existing objects out there! I can support the old metadata if needed.

### DIFF
--- a/api/python/t4/data_transfer.py
+++ b/api/python/t4/data_transfer.py
@@ -23,12 +23,12 @@ from .util import QuiltException, parse_file_url, parse_s3_url
 from . import xattr
 
 
-HELIUM_METADATA = 'helium'
-HELIUM_XATTR = 'com.quiltdata.helium'
+T4_METADATA = 't4'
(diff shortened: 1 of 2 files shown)


akarve:
@ResidentMario will this break Rob's upcoming notebook? hurdat?

akarve:
Can you create an alias in the enum (i.e. that would continue to support `PYARROW`)?
I am fine with this change but want to note that T4 will probably be rolled into a bigger project so not sure how useful it is to rename; but if it makes the code more readable, do it.

ResidentMario:
@akarve It looks like accessing old T4 objects in the new API would break, which is a background change that doesn't affect anything because we always access newly created objects in all of our tutorial etc. materials.

akarve:
Decided to wait on this.

PR #128, opened 2018-11-18
Fix the registry kwarg

dimaryaz:


akarve:
I think we should change the kwarg in line 255 (above) from ~`dest_registry`~ to `registry` as well.

dimaryaz:
And have two `registry` kwargs?

PR #129, opened 2018-11-18
Add a test to make sure we don't accidentally break top_hash()

dimaryaz:


api/python/tests/integration/test_packages.py
```diff
@@ -512,3 +512,13 @@ def test_diff():
...
+
+def test_top_hash_stable():
+    """Ensure that top_hash() never changes for a given manifest"""
+
+    registry = Path(__file__).parent / 'data'
+    pkg_hash = '20de5433549a4db332a11d8d64b934a82bdea8f144b4aecd901e7d4134f8e733'
+
+    pkg = Package.browse(registry=registry, pkg_hash=pkg_hash)
+
+    assert pkg.top_hash() == pkg_hash
```
akarve: good idea. i broke this test on purpose and the failure message wasn't that helpfu, so suggest adding the string:
`"Unexected tophash for data/{}".format(pkg_hash)"` or s/th like that

PR #130, opened 2018-11-18
Stop storing top_hash in the manifest

dimaryaz:
It's not actually used for anything... and it shouldn't be - we just calculate it as needed.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -424,8 +424,7 @@ class Package(object):
         """
         reader = jsonlines.Reader(readable_file)
         meta = reader.read()
-        # Pop the top hash -- it should only be calculated dynamically
-        meta.pop('top_hash', None)
+        meta.pop('top_hash', None)  # Obsolete
         pkg = cls()
         pkg._meta = meta
         for obj in reader:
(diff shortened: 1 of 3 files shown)


api/python/t4/packages.py
```diff
@@ -424,8 +424,7 @@ def load(cls, readable_file):
         """
         reader = jsonlines.Reader(readable_file)
         meta = reader.read()
-        # Pop the top hash -- it should only be calculated dynamically
-        meta.pop('top_hash', None)
+        meta.pop('top_hash', None)  # Obsolete
```
akarve: i guess we're keeping this in for old manifests? i would note as of which PR this becomes obsolete

api/python/tests/integration/data/local_manifest.jsonl
```diff
@@ -1,4 +1,4 @@
-{"semver": "1.1.1","comment": "example schema","top_hash": {"alg": "v0", "value": "c65a7b673b106c8051bd04db0976e84565dd1b5fc398f25b51eed79d2ff8bbfd"},"quilt_schema_version": "0.0.1"}
+{"semver": "1.1.1","comment": "example schema","quilt_schema_version": "0.0.1"}
```
akarve: consider bumping `quilt_schema_version` : we should have a README where we document changes so that it's easy to write migrations when we need to.

PR #132, opened 2018-11-18
Local registry package delete

ResidentMario:
Adds a top-level `t4.delete_package` method which allows deleting a package from a registry. Only works on a local registry, for now.


api/python/t4/api.py
```diff
@@ -101,6 +102,44 @@ def delete(target):
     delete_object(bucket, path)
 
 
+def delete_package(name, registry=None):
+    if not re.match(PACKAGE_NAME_FORMAT, name):
+        raise QuiltException("Invalid package name, must contain exactly one /.")
```
akarve: the RE could fail for other reasons than `/`, right?

ResidentMario: I don't think so. This check is a direct carry-over of `Package.validate_package_name`. It'd actually be better to move `validate_package_name` to `utils`, though, and just use that method here. TODO.

api/python/t4/api.py
```diff
@@ -101,6 +102,44 @@ def delete(target):
...
+        raise QuiltException("No such package exists in the given directory.")
+
+    registry_base_path = get_package_registry(fix_url(registry) if registry else None)
+    registry_url = urlparse(registry_base_path)
+    if registry_url.scheme != 'file':
+        raise NotImplementedError
+
+    pkg_namespace, pkg_subname = name.split("/")
+
+    registry_dir = pathlib.Path(parse_file_url(registry_url)).as_posix()
```
dimaryaz: Don't want the `as_posix()`

dimaryaz: `pathlib.Path` is not needed here - `registry_dir` is already a `Path`. Same below.

api/python/t4/api.py
```diff
@@ -101,6 +102,44 @@ def delete(target):
...
+    pkg_namespace_dir = pathlib.Path(registry_dir) / 'named_packages/' / pkg_namespace
+    pkg_dir = pkg_namespace_dir / pkg_subname
+    packages_path = pathlib.Path(registry_dir) / 'packages'
+
+    relevant_tophashes = []
+    for tophash_file in pkg_dir.glob('*/'):
+        # skip latest, which always duplicates a tophashed file
+        if tophash_file.name != 'latest':
+            with open(tophash_file, 'r') as f:
```
dimaryaz: `relevant_hashes.append(tophash_file.read_text())`

api/python/t4/api.py
```diff
@@ -101,6 +102,44 @@ def delete(target):
...
+
+    pkg_namespace, pkg_subname = name.split("/")
+
+    registry_dir = pathlib.Path(parse_file_url(registry_url)).as_posix()
+    pkg_namespace_dir = pathlib.Path(registry_dir) / 'named_packages/' / pkg_namespace
+    pkg_dir = pkg_namespace_dir / pkg_subname
+    packages_path = pathlib.Path(registry_dir) / 'packages'
+
+    relevant_tophashes = []
+    for tophash_file in pkg_dir.glob('*/'):
```
dimaryaz: `pkg_dir.iterdir()`

api/python/t4/api.py
```diff
@@ -101,6 +102,44 @@ def delete(target):
...
+        if tophash_file.name != 'latest':
+            with open(tophash_file, 'r') as f:
+                relevant_tophashes.append(f.read())
+
+        # remove the file
+        tophash_file.unlink()
+
+    pkg_dir.rmdir()
+
+    if len(list(pkg_namespace_dir.glob("*/"))) == 0:
```
dimaryaz: `if not list(pkg_namespace_dir.iterdir()):`

api/python/t4/api.py
```diff
@@ -101,6 +102,44 @@ def delete(target):
...
+        pkg_namespace_dir.rmdir()
+
+    for relevant_tophash in relevant_tophashes:
+        (packages_path / relevant_tophash).unlink()
```
dimaryaz: Buuut... how do you know some other package doesn't have the same hash? Or more like, the same package was built twice with different names.

api/python/tests/integration/test_packages.py
```diff
@@ -523,3 +523,13 @@ def test_top_hash_stable():
...
+
+
+@patch('appdirs.user_data_dir', lambda x, y: os.path.join('test_appdir', x))
+def test_local_package_delete(tmpdir):
+    """Verify that local package delete works."""
+    top_hash = Package().build("Quilt/Test")
+    t4.delete_package('Quilt/Test', registry=BASE_PATH)
+
+    assert 'Quilt/Test' not in t4.list_packages()
+    assert top_hash not in os.listdir(Path(BASE_PATH, '.quilt/packages').as_posix())
```
dimaryaz: Don't want `as_posix()`

ResidentMario:
@dimaryaz You should review again. I ended up on the check-everything path and implemented a helper method to help do that, `_tophashes_with_packages`.

api/python/t4/api.py
ResidentMario: Backlogged this method move. It's better addressed in its own PR as it's pretty merge conflict-y.

ResidentMario:
(the conflicts are due to colliding test updates, will resolve right before merge)

api/python/t4/api.py
```diff
@@ -101,6 +102,70 @@ def delete(target):
     delete_object(bucket, path)
 
 
+def _tophashes_with_packages(registry=None):
```
meffij: Please add a docstring so I can tell what this is supposed to do

api/python/t4/api.py
```diff
@@ -101,6 +102,70 @@ def delete(target):
...
+            for tophash in package_tophashes:
+                if tophash in out:
+                    out[tophash].append(pkg_name)
+                else:
+                    out[tophash] = [pkg_name]
+
+    return out
+
+
+def delete_package(name, registry=None):
```
meffij: definitely needs a docstring

meffij: There should be a summary of what all there is to be deleted with confirmation from the user, with a force=False argument.

meffij: maybe call it skip_confirmation

api/python/t4/api.py
```diff
@@ -101,6 +102,70 @@ def delete(target):
...
+
+    registry_base_path = get_package_registry(fix_url(registry) if registry else None)
+    registry_url = urlparse(registry_base_path)
+    if registry_url.scheme != 'file':
+        raise NotImplementedError
+
+    pkg_namespace, pkg_subname = name.split("/")
+
+    registry_dir = pathlib.Path(parse_file_url(registry_url))
+    pkg_namespace_dir = registry_dir / 'named_packages/' / pkg_namespace
```
meffij: why is there a `/` in `'named_packages/'`?

api/python/t4/api.py
ResidentMario: That's a good idea in spirit but nothing in the API right now takes user confirmation, and I don't really want to start doing that here right now.

api/python/t4/api.py
ResidentMario: Good catch. TIL `pathlib.Path` handles this for you:

```python
'a' / 'b' == 'a/' / 'b' 
```

api/python/t4/api.py
ResidentMario: E.g. I think we should also consider having confirmation on `delete` and `delete_path` (which is a TODO), and that it should all be considered in a separate PR.

meffij: I feel pretty strongly that we shouldn't ship this without a confirmation. @akarve please weigh in

ResidentMario: Well, we already don't ask for confirmation on `delete`.

ResidentMario: Can we leave that out of this PR and open a PR after this one for confirmation messages on `{delete, delete_dir, delete_package}`?

meffij: I'm OK with that

PR #134, opened 2018-11-19
Implement S3 directories in set_dir

dimaryaz:


api/python/t4/packages.py
```diff
@@ -461,10 +464,21 @@ def set_dir(self, lkey, path):
...
                 logical_key = lkey + f.relative_to(src_path).as_posix()
                 # TODO: Warn if overwritting a logical key?
                 self.set(logical_key, entry)
+        elif url.scheme == 's3':
+            src_bucket, src_key, src_version = parse_s3_url(url)
+            if src_version:
+                raise PackageException("Directories cannot have versions")
+            objects = list_objects(src_bucket, src_key)
+            for obj in objects:
+                obj_url = 's3://%s/%s' % (src_bucket, quote(obj['Key']))
```
meffij: what about version ID?

api/python/t4/packages.py
```diff
@@ -461,10 +464,21 @@ def set_dir(self, lkey, path):
...
+                entry = PackageEntry([obj_url], None, None, None)
+                logical_key = lkey + obj['Key'][len(src_key) + 1:]
```
meffij: Is there any way to make this line more clear? I'm not sure why you're doing it that way

api/python/t4/packages.py
dimaryaz: Ooh, yes

api/python/t4/packages.py
dimaryaz: Don't know if there is a way - we have to strip the prefix.

E.g., if I call `set_dir('foo', 'x/y')`, S3 will give me keys like:
```
x/y/a.txt
x/y/b.txt
```
and I need to turn them into:
```
foo/a.txt
foo/b.txt
```
So not much to do other than substring?

PR #135, opened 2018-11-19
Create closure so metadata gets set correctly

meffij:
Helpful article: http://stupidpythonideas.blogspot.com/2016/01/for-each-loops-should-define-new.html


meffij:
Another way to express this idiom is 
```
def meta_callback(resp, _key=key):
  meta = _parse_metadata(resp)
  with lock:
    metadata[_key] = meta
```
@dimaryaz which do you prefer?

meffij:
It applies to every language that re-uses a variable binding during loops. Python, Ruby, JS all require this pattern if you want closures in loops.

PR #138, opened 2018-11-20
Better package repr

meffij:
@ResidentMario do we want to include any other data? Right now it's just node names


api/python/t4/packages.py
```diff
@@ -230,6 +230,18 @@ def __init__(self):
...
+    def __repr__(self, level=0, indent='  '):
+        """
+        String representation of the Package.
+        """
+        self_repr = ''
+        for child_key in sorted(self.keys()):
+            child_entry = indent*level + child_key + '\n'
+            self_repr += child_entry
+            if isinstance(self[child_key], Package):
+                self_repr += self[child_key].__repr__(level+1, indent)
```
stevededalus: You could save yourself an argument by just appending to indent when making the recursive call (no need for level then). Strings are immutable in python so that won't impact the string in the calling frame
```
self_repr += self[child_key].__repr__(constants.INDENT + indent)
```
No strong preference, just pointing out an option.

meffij: I like having it as an argument in case someone wants to use a different indent. It's a very small consideration, but it's why I made it an argument instead of a constant.

meffij:
If the max length is 20 lines, and a package has 25 top-level directories, is it correct to show no grandchildren, 19 child directories and the `...`, or should we show some grandchildren? What is the correct behavior to cut the line count down?

ResidentMario:
In that case I think that `19 child directories and the ...` would be best. If the user wants to look at individual files in some directory, they can key into the package directory they're interested in, get the subpackage `repr`, and see those files then. But if we don't show them the directory they have access to right in the `repr`, they won't know where to go to get those files, so their ability to proceed would be hampered (they would have to print out `pkg.keys()` and look through that). So I think showing as many top-level folders as possible is higher priority than showing many files.

meffij:
@ResidentMario Try out this new version. Also, should we make `max_lines` higher than 20?

ResidentMario:
20 is about how many will fit in a terminal window (at least on my machine) so it seems like a good starting point. The output of this implementation looks fine. One problem spot: printing an empty Package results in an empty string. For now that's fine, but we might want to tweak that later.

PR #140, opened 2018-11-21
Make Package.dump work without hashes

dimaryaz:
Just serialize the package; don't check if the hashes are set.


meffij:
Why are we doing this? Is this just temporary until we have a PackageBuilder?

PR #143, opened 2018-11-22
map, filter, and reduce

ResidentMario:
These operators are designed to enable munging packages on an element-by-element basis. They address the most common use-case for processing package metadata: iterating over it.

They are not scoped for dealing with metadata dependencies, so are less useful when the user has package folders above the entry which contain their own metadata, and need to use that metadata to inform how they parse the package. To do that the user will need to parse the package themselves. We accepted this tradeoff when we decided to allow setting metadata at the directory level.


api/python/t4/packages.py
```diff
@@ -841,3 +841,27 @@ def diff(self, other_pkg):
...
+        return [(lk, entity) for lk, entity in self.walk() if f(lk, entity)]
+
+    def reduce(self, f, default=None, include_directories=False):
+        if include_directories:
+            raise NotImplementedError
+
+        entities = list(self.walk())
+        out = default if default is not None else entities[0]
+
+        for i in range(len(entities)):
```
stevededalus: I think you're double counting entries[0] of there's no default.
Also you can probably use a straight enumeration instead of `range(len())`

api/python/t4/packages.py
```diff
@@ -841,3 +841,27 @@ def diff(self, other_pkg):
         added = list(sorted(other_entries))
         
         return added, modified, deleted
+
+    def map(self, f, include_directories=False):
```
stevededalus: docstrings?

api/python/tests/integration/test_packages.py
```diff
@@ -623,3 +623,28 @@ def test_manifest():
...
+    pkg.set('as/qw', LOCAL_MANIFEST)
+    assert pkg.filter(lambda lk, entry: lk == 'as/df') == [('as/df', pkg['as/df'])]
+
+
+def test_reduce():
+    pkg = Package()
+    pkg.set('as/df', LOCAL_MANIFEST)
+    pkg.set('as/qw', LOCAL_MANIFEST)
+    assert pkg.reduce(lambda a, b: a) == ('as/df', pkg['as/df'])
+    assert pkg.reduce(lambda a, b: b) == ('as/qw', pkg['as/qw'])
```
stevededalus: This seems like a werid implementation. I would expect `lambda a, b: a` in a reduce to return the first element only of a list, not a tuple.

ResidentMario: Entries are not aware of their logical key, but the logical key is very useful for any operation you will perform on an entry. So the options were:

* Omit the logical key: `(a, b) = left_entry, right_entry`
* Include the logical key and make the reducer operate on tuples: `(a, b) = (left_lk, left_entry), (right_lk, right_entry)`
* Include the logical keys as separate arguments: `(a_1, a_2, b_1, b_2) = (left_lk, left_entry, right_lk, right_entry)`.

And I chose the second option. What do you think?

api/python/t4/packages.py
ResidentMario: Docstrings added.

api/python/t4/packages.py
ResidentMario: Good catch. New behavior:

        entities = list(self.walk())
        out = default if default is not None else entities[0]
        iters = entities if default is not None else entities[1:]

        for entry in iters:
            out = f(out, entry)

api/python/t4/packages.py
```diff
@@ -843,25 +843,75 @@ def diff(self, other_pkg):
...
+
+        Args:
+            f: function
+                The function to be applied to each package entry.
+                It should take two inputs, a logical key and a PackageEntry.
+            include_directories: bool
+                Whether or not to include directory entries in the map.
+
+        Returns: list
+            The list of results generated by the map.
```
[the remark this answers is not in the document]
ResidentMario: Good point. It wouldn't be hard to do, but for right now I'm going to put off doing so in favor of getting this PR merged.

PR #149, opened 2018-11-28
Fall back to unsigned boto client if user doesn't have valid AWS credentials

meffij:
`alpha-quilt-storage` is a placeholder for some public bucket that you should be able to HEAD. We should replace this with something that makes sense, suggestions welcome


meffij:
Changing to `quilt-example` since that bucket is intended to be publically readable

PR #154, opened 2018-12-04
Bucket.search

meffij:
Tests in progress

### DIFF
--- a/api/python/t4/bucket.py
+++ b/api/python/t4/bucket.py
@@ -1,3 +1,9 @@
+"""
+bucket.py
+
+Contains the Bucket class, which provides several useful functions
+    over an s3 bucket.
+"""
 import json
 import pathlib
 from urllib.parse import urlparse
@@ -7,6 +13,7 @@ import requests
 from .data_transfer import (TargetType, copy_file, copy_object, delete_object,
                             deserialize_obj, get_bytes, get_size_and_meta,
                             list_objects, put_bytes, serialize_obj)
+from .search import search
 from .util import QuiltException, fix_url, parse_s3_url
 
 CONFIG_URL = "https://t4.quiltdata.com/config.json"
@@ -50,15 +57,29 @@ class Bucket(object):
                 raise QuiltException("Config at config_url malformed")
             return
         if self._bucket in configs:
-            self._search_endpoint = configs[self._bucket]
+            self._search_endpoint = configs[self._bucket]['search_endpoint']
         elif not quiet:
             raise QuiltException("Config info not found for this bucket")
 
     def search(self, query):
         """
-        Placeholder till I decide how to move the logic around
+        Execute a search against the configured search endpoint.
+
+        query: query string to search
+
+        Returns either the request object (in case of an error) or
+                a list of objects with the following keys:
+            key: key of the object
+            version_id: version_id of object version
+            operation: Create or Delete
+            meta: metadata attached to object
+            size: size of object in bytes
+            text: indexed text of object
+            source: source document for object (what is actually stored in ElasticSeach)
+            time: timestamp for operation
+
         """
-        pass
+        return search(query, self._search_endpoint)
 
     def deserialize(self, key):
         """
@@ -228,7 +249,6 @@ class Bucket(object):
         Raises:
             if put to bucket fails
         """
-        dest = self._uri + key
         existing_meta = self.get_meta(key)
         existing_meta['user_meta'] = meta
         copy_object(self._bucket, key, self._bucket, key, existing_meta)
--- a/api/python/t4/bucket.py
+++ b/api/python/t4/bucket.py
@@ -1,10 +1,15 @@
+import json
 import pathlib
 from urllib.parse import urlparse
+
+import requests
 
 from .data_transfer import (TargetType, copy_file, copy_object, delete_object,
                             deserialize_obj, get_bytes, get_size_and_meta,
                             list_objects, put_bytes, serialize_obj)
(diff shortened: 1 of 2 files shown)


api/python/t4/bucket.py
```diff
@@ -27,6 +39,47 @@ def __init__(self, bucket_uri):
...
+            version_id: version_id of object version
+            operation: Create or Delete
+            meta: metadata attached to object
+            size: size of object in bytes
+            text: indexed text of object
+            source: source document for object (what is actually stored in ElasticSeach)
+            time: timestamp for operation
+
+        """
+        return search(query, self._search_endpoint)
```
ResidentMario: We're silently placing the user in a scenario where the bucket they are interacting with may or may not be searchable. If they later go to try to search that bucket, and that bucket isn't searchable, they get some obscure error---frustrating.

We should throw a better error when search fails because the bucket isn't properly configured for it.

api/python/t4/search_util.py
```diff
@@ -0,0 +1,86 @@
...
+                'meta': meta,
+                'size': size,
+                'text': text,
+                'source': result['_source'],
+                'time': time
+            })
+        results = list(sorted(results, key=lambda x: x['time'], reverse=True))
+        return results
+    except KeyError:
+        return raw_response
```
ResidentMario: Why would returning a raw response object in this case be useful to the end user? If my function fails I'd generally expect to get an error back, not a blob.

api/python/tests/test_search.py
```diff
@@ -0,0 +1,54 @@
...
+                        'type': 'asdf',
+                        'user_meta': {},
+                        'size': 0,
+                        'text': ''
+                    }
+                }]
+            }
+        }
+        create_es_mock.return_value = es_mock
+        b = Bucket('s3://quilt-example')
```
ResidentMario: We use `s3://test-bucket` as our mock bucket elsewhere, should do here too for consistency.

api/python/t4/search_util.py
meffij: no good reason; it's more helpful to the user to see whatever Elasticsearch says their problem was than to see a generic error from us

ResidentMario: I doubt most users will even know what an Elasticsearch is, let alone that that's what we're using, much better here to be consistent and up-front about the error.

api/python/t4/bucket.py
meffij: I made the error messages a little better, what do you think?

api/python/t4/search_util.py
meffij: What verbiage would you recommend?

api/python/t4/bucket.py
ResidentMario: Current error message:

> QuiltException: Config info not found for this bucket at config_url

I would shorten this to just:

> QuiltException: Config info not found for this bucket

"at `config_url`" implies to me that there is `Bucket.config_url` attribute that I can access to further debug the issue, when in reality no such attribute exists, which is unhelpful.

Once we have documentation we'll need a section for FAQs or common issues that contains an entry for this config problem, which we'll probably want to point the user to from this error message. Task for the future.

api/python/t4/search_util.py
```diff
@@ -0,0 +1,86 @@
...
+
+    raw_response = es_client.search(index=ES_INDEX, body=payload)
+
+    try:
+        results = []
+        for result in raw_response['hits']['hits']:
+            key = result['_source']['key']
+            vid = result['_source']['version_id']
+            operation = result['_source']['type']
+            meta = json.dumps(result['_source']['user_meta'])
```
stevededalus: Is this field guaranteed to be set?

api/python/t4/search_util.py
stevededalus: Seems like it should raise an exception that contains an error message with the raw response inside it?

ResidentMario: I'd throw `QuiltError("Query failed unexpectedly due to either a bad query or a misconfigured search service.")` (or similar).

ResidentMario: @stevededalus I've never patched an `Exception` to provide a failing response as a param (is that what you are suggesting?), though in retrospect [obviously](https://stackoverflow.com/questions/16466406/passing-an-object-with-an-exception) you can do that. Is this something that one would reasonably expect from an API?

stevededalus: No, but it's useful in `alpha` for us to debug customer issues.

api/python/t4/search_util.py
meffij: yes it will be `{}` if the indexer doesn't see any user meta

api/python/t4/search_util.py
```diff
@@ -0,0 +1,86 @@
...
+def search(query, search_endpoint, aws_region='us-east-1'):
+    """
+    Searches your bucket. query can contain plaintext, and can also contain clauses
+    like $key:"$value" that search for exact matches on specific keys.
+
+    Returns either the request object (in case of an error)
+            or a list of objects with the following keys:
+        key: key of the object
+        version_id: version_id of object version
+        operation: Create or Delete
```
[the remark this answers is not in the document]
meffij: It's either `"Create"` or `"Delete"` -- when you delete something, we insert a document into ES with barebones data and the operation type "Delete"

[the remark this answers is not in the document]
stevededalus: So this is a tombstone object that signifies that the entry is deleted?

PR #162, opened 2018-12-07
Updated T4 CloudFormation template for publicly searchable T4 buckets

meffij:
- contains an example bucket policy for making your s3 bucket publicly readable and listable

- adds a condition to allow public t4 search

- serves everything with the `configs/` prefix in your config bucket through CloudFront


deployment/t4.yaml
```diff
@@ -20,6 +20,15 @@ Parameters:
   Users:
     Type: CommaDelimitedList
     Description: Required, comma-separated list of ARNs for IAM users who can search QuiltBucket. NOTE - Users with search access can see the contents of all indexed documents. Only include trusted users.
+  PublicHostedBucket:
+    Type: String
```
akarve: so `Bool` is not a thing?

akarve: Are we sure we want to do this in CFN at all? Public buckets is *only* a thing for hosted buckets, and in that case, I don't think we want to run a CFN template every time someone buys a bucket.

PR #163, opened 2018-12-07
Update Bucket.config to new schema

meffij:


stevededalus:
Can you drop a link to the config file in S3 that its reading?

meffij:
https://s3.amazonaws.com/alpha-quilt-config/config.json

meffij:
https://s3.amazonaws.com/t4-config-bucket/config.json

it's not updated to the new format yet, which is why this PR won't work correctly yet

PR #164, opened 2018-12-10
Catalog: config.json + federations

nl0:
Changes:

* read config from `/config.json`, remove env-based config logic
* upgrade config schema
* use injectable async config throughout the app
* cleanup obsolete stuff from `containers/App`
* adjust and refactor stuff in `containers/Bucket` and `containers/NavBar` to support new config injection mechanism
* `utils/BucketConfig` module
* `utils/Data`: refactor to use render prop, fixing some bugs along the way
* `utils/Delay` helper component
* `utils/Wait` helper module
* `utils/validate` helper module
* refactor `SagaInjector` and `ReducerInjector`


catalog/app/constants/config.js
```diff
@@ -1,39 +1,30 @@
...
-
-if (!check(config)) {
-  throw new Error(`Invalid config:\n${JSON.stringify(config, null, 2)}`);
-}
+const check = R.unless(
+  conforms({
+    alwaysRequiresAuth: R.is(Boolean),
+    sentryDSN: isNullable(String),
+    apiGatewayUrl: R.is(String),
+    defaultBucket: R.is(String),
```
meffij: why do we need defaultBucket?

catalog/config.json.example
```diff
@@ -0,0 +1,29 @@
+{
+  "alwaysRequiresAuth": false,
+  "sentryDSN": "<DSN here>",
+  "apiGatewayUrl": "https://ozj6r9bglc.execute-api.us-east-1.amazonaws.com/prod",
```
meffij: please do the same as with sentry, replace with `<apiGateway url here>` or similar

catalog/app/constants/config.js
nl0: to test if credentials are valid and to redirect after sign-in

catalog/app/constants/config.js
```diff
@@ -1,39 +1,30 @@
...
+    buckets: R.both(R.is(Array), R.all(conforms({
+      name: R.is(String),
+      title: isNullable(String),
+      icon: isNullable(String),
+      description: isNullable(String),
+      searchEndpoint: isNullable(String),
+      menu: isNullable(Boolean),
```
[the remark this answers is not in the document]
nl0: in this context "nullable" and "optional" mean the same, the prop can be omitted or set to `null` -- there's no difference

nl0:
@meffij i've started working on this before i saw the spec, so this logic will be reimplemented to adhere to the spec

meffij:
Deployed hash 836166e on `t4-stage.quiltdata.com`

nl0:
@meffij FYI: this is still WIP

nl0:
@meffij @akarve this one seems ready, have a look please

catalog/app/utils/Config.js
```diff
@@ -0,0 +1,173 @@
...
+const parseJSON =
+  R.tryCatch(JSON.parse, (e, src) => {
+    throw new ConfigError(`invalid JSON:\n${src}`, { src, originalError: e });
+  });
+
+const validateConfig = conforms({
+  alwaysRequiresAuth: R.is(Boolean),
+  sentryDSN: isNullable(String),
+  apiGatewayEndpoint: R.is(String),
+  defaultBucket: R.is(String),
```
nl0: @meffij the idea of using listBuckets didnt work out: this method doesn't support CORS, so we're back to listing the default bucket contents.
also, i've added explicit sign-in / sign-out redirects to the config

PR #167, opened 2018-12-11
Remote package delete

ResidentMario:
Implements deleting packages from remote S3 stores. Follow-up to #132, which enabled local package delete.

This PR also contains a correction to the tophash reference counting code for local packages. Previously it was incorrectly counting packages which point to the same _timestamp_, instead of the same _tophash_.


meffij:
If you merge master then your codecov will be updated

ResidentMario:
Rebased onto current master and looks like codecov is green again.

PR #168, opened 2018-12-11
Pypi ci

stevededalus:
Add a circleci workflow to update PyPi backages based on git tag commands. Added a function to verify that the git tag matches the pypi version in `setup.py`

To cut a release:

1. Update the version in `setup.py`
2. `git tag $VERSION`
3. `git push --tags`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,4 +99,5 @@ jobs:
           name: upload to pypi
           command: |
             . venv/bin/activate
+            pip install twine
             twine upload dist/*
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ workflows:
             - build
           filters:
             tags:
-              only: /[0-9]+(\.[0-9]+)*/
+              only: /[0-9]+(\.[0-9]+)*(\-[a-ZA-Z])*/
             branches:
               ignore: /.*/
 jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,32 @@ jobs:
             . venv/bin/activate
             cd api/python && pytest --cov=./
             codecov
+  build:
+    docker:
+      - image: circleci/python:3.6-jessie
+    steps:
+      - checkout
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -24,3 +35,69 @@ jobs:
...
+          command: |
+            python3 -m venv venv
+            . venv/bin/activate
+            python api/python/setup.py verify
+      - run:
+          name: init .pypirc
+          command: |
+            echo -e "[pypi]" >> ~/.pypirc
+            echo -e "username = quiltdata" >> ~/.pypirc
+            echo -e "password = $PYPI_PASSWORD" >> ~/.pypirc
```
akarve: are we sure this doesn't end up in the logs? a safer way of doing this might be `envsubst` on a template file.

stevededalus: Do all this logs is:

```
#!/bin/bash -eo pipefail
echo -e "[pypi]" >> ~/.pypirc
echo -e "username = quiltdata" >> ~/.pypirc
echo -e "password = $PYPI_PASSWORD" >> ~/.pypirc
```

I don't think `envsubst` is any safer.

api/python/setup.py
```diff
@@ -1,13 +1,38 @@
...
+
+VERSION = "0.0.3"
 
 def readme():
     readme_short = """
+    Quilt T4 is a data management tool designed for data discoverability, data dependency
+    management, and data version control using `data packages <https://blog.quiltdata.com/data-packages-for-fast-reproducible-python-analysis-c74b78015c7f>`_.
+
+    The `t4` PyPi package allows you to build, push, and pull data packages in T4 using Python.
+    Visit the GitHub `repository <https://github.com/quiltdata/t4>`_ for more information.
```
ResidentMario: Now that docs are up we can update this link text to:

```
Visit the `documentation quickstart <https://quiltdocs.gitbook.io/t4/quickstart>`_ for more information.
```

stevededalus: Done, thanks

PR #173, opened 2018-12-12
Repair catalog build script

ResidentMario:
The catalog build script as it exists on the current `master` branch of this repository fails for me. This occurs due to two separate but relatedly-located reasons.

The first is that the webpack builder relies on the `jpegtran-bin` package to perform certain types of image compression. However, the version of `jpegtran-bin` that is installed via a clean `npm install` as of the current `master` is missing vendor components:

```
$ cd catalog
$ ls node_modules/jpegtran-bin

cli.js		lib		node_modules	readme.md
index.js	license		package.json	test
```

The fix is to include `jpegtran-bin` in`devDependencies` explicitly. Now when you perform a clean `npm install` the vendor-ed components are included:

```
$ cd catalog
$ ls node_modules/jpegtran-bin

cli.js		lib		node_modules	readme.md	vendor
index.js	license		package.json	test
```

The second issue is less clear to me. The `image-webpack-loader` loader, as configured on current `master`, is resulting in build failures for me. Specifically, the following setting in `catalog/internal/webpack.base.babel.js` is troublesome:

```js
      {	
        test: /\.(jpg|png|gif)$/,	
        use: [	
          'file-loader',	
          {	
            loader: 'image-webpack-loader',	
            options: {	
              progressive: true,	
              optimizationLevel: 7,	
              interlaced: false,	
              pngquant: {	
                quality: '65-90',	
                speed: 4,	
              },	
            },	
          },	
        ],	
      }
```

I backed this out to the default `file-loader` which IIRC is used by the base `react-boilerplate-quickstart`:

```js
{
     test: /\.(eot|svg|otf|ttf|woff|woff2|jpg|png|gif)$/,
     use: 'file-loader'
}
```

Since we're losing compression we may be dropping page load speed here. But this at least builds. What it is about the `image-webpack-loader` that fails on a clean install remains to be investigated.

### DIFF
--- a/catalog/internals/webpack/webpack.base.babel.js
+++ b/catalog/internals/webpack/webpack.base.babel.js
@@ -47,26 +47,8 @@ module.exports = (options) => ({
         use: ['style-loader', 'css-loader'],
       },
       {
-        test: /\.(eot|svg|otf|ttf|woff|woff2)$/,
-        use: 'file-loader',
-      },
-      {
-        test: /\.(jpg|png|gif)$/,
-        use: [
-          'file-loader',
-          {
-            loader: 'image-webpack-loader',
-            options: {
-              progressive: true,
-              optimizationLevel: 7,
-              interlaced: false,
-              pngquant: {
-                quality: '65-90',
-                speed: 4,
-              },
-            },
-          },
-        ],
+        test: /\.(eot|svg|otf|ttf|woff|woff2|jpg|png|gif)$/,
+        use: 'file-loader'
       },
       {
         test: /\.html$/,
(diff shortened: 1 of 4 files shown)


ResidentMario:
Per @akarve before we merge we should figure out what the issue with `image-webpack-loader` is and try and get it working again. @nl0 do you have any ideas?

nl0:
@ResidentMario `image-webpack-loader` is seriously outdated, so we should probably remove it altogether. we don't have that many images in our bundle, and i think we can optimize them and newly added ones by hand.

ResidentMario:
Per @akarve we can merge this PR, with consideration for documenting the need to only check optimized image files into the repository in `CONTRIBUTING.md`.

nl0:
we should remove obsolete packages

ResidentMario:
Pruning obsolete packages like `image-webpack-loader` from the catalog `package.json` would be a good idea. But I'm not aware of what is and isn't used by the catalog, so you're much better placed to do that. 🙂

nl0:
@ResidentMario cool, i've backlogged it

PR #177, opened 2018-12-12
Add markdown to indexed plaintext

meffij:


lambdas/es/indexer/lam.py
```diff
@@ -85,7 +85,9 @@ def format_notebook(nb):
     codecells = filter(lambda x: x['cell_type'] == 'code', cells)
     formatted = map(format_codecell, codecells)
     text = ' '.join(list(formatted))
-    return text
+    markdowncells = filter(lambda x: x['cell_type'] == 'markdown', cells)
+    markdowntext = ' '.join([' '.join(cell['source']) for cell in markdowncells])
+    return text + ' ' + markdowntext
```
akarve: suggest string join something other than space - maybe a comment that explains what the section break dows

meffij: The problem with this is whatever separator we use will be indexed. For example, if we join on `" Markdown: "` then every notebook will match on `markdown`

PR #178, opened 2018-12-13
Make package physical keys a singleton

ResidentMario:
Per discussion ([circa one month ago](https://quiltdata.slack.com/archives/C3A5VP2MP/p1542241721023800)) we agreed that package physical keys, which are currently a list, should become a singleton.

This PR refactors the package manifest format to match this new pattern. It also bumps the `"version": "v0"` reference at the top of the manifest to `"version": "v1"`.

This PR does _not_ do something else that was also proposed, but also still needs to be implemented: hanging the package metadata off of a `'/'` key in the manifest file, instead of having it live on its own special line at the top of the manifest file. That is left for a future PR.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -34,24 +34,6 @@ def hash_file(readable_file):
 
     return hasher.hexdigest()
 
-def _to_singleton(physical_keys):
-    """
-    Ensure that there is a single physical key, throw otherwise.
-    Temporary utility method to avoid repeated, identical checks.
-
-    Args:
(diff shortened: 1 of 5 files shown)


akarve:
Nice initiative. I _did_ say that the current design was suboptimal. Nevertheless, with Parquet support on the horizon (where a single table may be broken into multiple files), and given the fact that this change would incur a schema migration, I suggest we close this PR and leave as-is until we get our heads wrapped around Parquet partitions + datasets.

PR #179, opened 2018-12-13
Update installation section

akarve:


akarve:
@ResidentMario note that these [indentations](https://github.com/quiltdata/t4/pull/179/files#diff-045d5d7186acf52b74848a7dfbdde200R132) of image blocks are necessary to preserve the hierarchy of the document.

docs/Installation.md
```diff
@@ -1,58 +1,135 @@
...
 ```bash
-$ pip install git+https://github.com/quiltdata/t4.git#subdirectory=api/python
+$ pip install aws-cli
+$ aws configure
 ```
 
-In order to interface with a remote catalog you will also need valid AWS credentials. If this is your first time using AWS, ensure that you have the AWS CLI installed (`pip install aws-cli`), then run the following to store the IAM credentials you wish to use with T4:
+If you are already using the AWS CLI, you may use your existing profile, or [create a new profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-multiple-profiles.html).
+
+### Developer
```
ResidentMario: If folks want to know how to do a developer install we should direct them to `CONTRIBUTING.md`, and not embed it into the text here.

docs/Installation.md
```diff
@@ -1,58 +1,135 @@
...
+
 T4 has two different components:
-* A Python package, `t4`, distributed via `pip`
-* A web interface, the T4 Catalog, deployed using [AWS CloudFormation](https://aws.amazon.com/cloudformation/)
+* A Python package, `t4`
+* A web interface, the T4 Catalog
 
 If you have an already-provisioned catalog, only the Python package is needed.
 
+## T4 client (Python 3.6)
```
ResidentMario: We should keep the title clean and state that you need Python >=3.6 somewhere in the text, not in the heading.

docs/Installation.md
```diff
@@ -1,58 +1,135 @@
...
 
 ```bash
-$ aws configure
+$ pip install git+https://github.com/quiltdata/t4.git#subdirectory=api/python
 ```
 
-If you are already using AWS IAM you may reuse your credentials for T4. Alternatively you can create a Quilt-specific [profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-multiple-profiles.html).
+## Catalog
 
+The Javascript catalog is a single-page, static application.
```
ResidentMario: Is the word static meaningful here? I think just "single page application" is the keyword you want.

docs/Installation.md
```diff
@@ -1,58 +1,135 @@
...
-* Go to Services > CloudFormation > Create stack
+For more advanced deployment scripts, see [`deploy_navigator.sh`](../deployment/deploy_navigator.sh).
 
-![](imgs/start.png)
 
-* Click "Upload a template to Amazon S3" and select `t4.yaml`, provided to you by Quilt
-* Click Next
-* Fill in Stack name and Parameters
+## AWS T4 Instance (via CloudFormation)
+The following section uses CloudFormation to manage S3 buckets, ElasticSearch endpoints, and CloudFront to deploy
```
ResidentMario: You need to explain what these components are, e.g. state that these instructions are for deploying a catalog instance on an AWS account.

docs/Installation.md
```diff
@@ -1,58 +1,135 @@
...
+a T4 instance in your VPC.
 
-![](imgs/params.png)
+### Known limitations
 
-> Carefully note parameter descriptions to avoid stack failure
+* Supports only one bucket
```
ResidentMario: Is this still true?

docs/Installation.md
akarve: Static is important because there is no webserver required

docs/Installation.md
akarve: Yes

docs/Installation.md
akarve: From experience with Quilt 2, it's useful to have this in the mainline docs, because users who are trying to work around a bug, or get early grips on a fix, will do just this. Agree with your comment for anything more complicated.

docs/Installation.md
akarve: Rather than that, I promised less.

PR #180, opened 2018-12-14
More helpful error for unexpected bucket URIs

akarve:


api/python/t4/util.py
```diff
@@ -88,7 +88,7 @@ def parse_s3_url(s3_url):
     Takes in the result of urlparse, and returns a tuple (bucket, path, version_id)
     """
     if s3_url.scheme != 's3' or not s3_url.netloc or (s3_url.path and not s3_url.path.startswith('/')):
-        raise ValueError("Malformed S3 URI")
+        raise ValueError("Expected URI scheme 's3://', not '{}'".format(s3_url.scheme))
```
ResidentMario: The correct `url_scheme` reported by this error string is `s3://`, but in the correct case actually `s3_url.scheme = "s3"`.

Is it possible to return `s3_url.scheme` plus whatever the leading characters the user entered were? E.g. if they entered `abcd://` they should get back `"Expected...'s3://', not 'abcd://'` (as opposed to what would be returned now, which is `"Expected...'s3://', not 'abcd'`).

akarve: Yeah, I've just fixed this by handling every case separately, which I should have done in the first place.

akarve:
Regarding TODO, I'm leaving as is to preserve the original author's intent. It's not to fix in this PR.

PR #182, opened 2018-12-15
Fix upload_file for large_files

kevinemoore:
## Description
Passing a callback to grab the VersionId works for single-part uploads (put_file), but not multi-part uploads. This PR adds a separate codepath for single-file uploads (i.e. not directories) that checks the file size and uses a multi-part upload for large files. It grabs the VersionId from `complete_multipart_upload`. There is likely an opportunity to speedup large-file uploads by parallelizing the uploading of the separate parts, but I'm saving that for a future PR.

### DIFF
--- a/api/python/t4/data_transfer.py
+++ b/api/python/t4/data_transfer.py
@@ -363,7 +363,7 @@ def upload_file(src_path, bucket, key, override_meta=None):
         src_etag_list = list(src_etag_iter)
 
     with tqdm(total=total_size, unit='B', unit_scale=True) as progress:
-        callback = ProgressCallback(progress)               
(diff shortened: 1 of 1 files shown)


stevededalus:
LGTM for fixing the current issue, but this is continuing the highlight the messiness of our datamodel wrt to guarantees of reproducibility.

I think that we should seriously consider dropping support for logical key directories. Since they can't be versioned, they produce different behavior for a package's immutability guarantees. Calling `set('/','/')` vs `for i in set: set(/+'i',/+'i')` produce packages that behave differently, one of which is reproducible because it attaches versioning, the other isn't.

PR #184, opened 2018-12-17
Add version info on set for s3 objects

stevededalus:
## Description

Add a version to s3 keys on `package.set` calls.

## TODO

- [x] Unit Tests

### DIFF
--- a/api/python/t4/data_transfer.py
+++ b/api/python/t4/data_transfer.py
@@ -595,9 +595,6 @@ def get_size_and_meta(src):
     """
     Gets metadata for the object at a given URL.
 
-    If no S3 versionId is present in the URL it will return a versionId for a current object.
-    If an S3 versionId is present in the input then the returned version will be none.
-
     Returns:
         size, meta(dict), version(str)
     """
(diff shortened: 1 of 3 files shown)


api/python/t4/data_transfer.py
```diff
@@ -594,8 +594,15 @@ def get_bytes(src):
 def get_size_and_meta(src):
     """
     Gets metadata for the object at a given URL.
+
+    If no S3 versionId is present in the URL it will return a versionId for a current object.
```
kevinemoore: This behavior seems strange to me. I'd expect this method to return the versionid regardless of the inputs. It might also be useful to return the `is_current` value.

api/python/t4/packages.py
```diff
@@ -737,7 +737,9 @@ def set(self, logical_key, entry, meta=None):
         """
         if isinstance(entry, (string_types, getattr(os, 'PathLike', str))):
             url = fix_url(str(entry))
-            size, orig_meta = get_size_and_meta(url)
+            size, orig_meta, version = get_size_and_meta(url)
+            if version:
```
kevinemoore: It would be clearer to just test the url here for the presence of a versionid.

api/python/t4/data_transfer.py
stevededalus: Yeah, I was trying to avoid parsing the same url twice to check the version id, but i agree that this is convoluted. Switched back to checking in both places.

PR #191, opened 2018-12-19
Improve text extraction for use by ElasticSearch

akarve:
## Description
* Unit tests
* Use `nbformat` for correctness + robustness
* Don't index noisy fields like streams


lambdas/es/indexer/lam.py
```diff
@@ -63,33 +61,46 @@ def transform_meta(meta):
...
+    code = []
+    markdown = []
+    for cell in formatted.get('cells', []):
+        if 'source' in cell and 'cell_type' in cell:
+            if cell['cell_type'] == 'code':
+                code.append(cell['source'])
+            elif cell['cell_type'] == 'markdown':
+                markdown.append(cell['source'])
+
+    return '\n'.join(code + markdown)
```
meffij: This does not keep the code and markdown in order -- it will do all the code, then all the markdown. Just use one output list, and put all code and markdown in it as you go through.

.circleci/config.yml
```diff
@@ -58,6 +59,7 @@ jobs:
           name: run tests
           command: |
             . venv/bin/activate
+            cd lambdas && pytest --cov=./ && cd ..
```
meffij: change to `--cov=./**.py` to avoid coverage report including test data (notebooks)

lambdas/es/indexer/lam.py
akarve: it's now in-order

PR #192, opened 2018-12-20
Move some top-level methods to the Bucket API

ResidentMario:
In general we want to shrink the surface area of the top-level API (`t4.*`) by moving methodology that exists there that would better live on the `Bucket` object (`t4.Bucket.*`). In this refactor I move three such methods: `delete`, `delete_dir`, and `ls`.


ResidentMario:
Our remaining top-level API surface:
* `copy` -- this is probably useful as a low-level API method for users, so I'd suggest keeping this for now.
* `put` -- duplicated by `t4.Bucket.put`, @akarve is this safe to remove?
* `get` -- duplicated and renamed by `t4.Bucket.deserialize`, @akarve is this safe to remove?
* `search` -- @meffij anything that prevents us from moving this to `Bucket`?
* `config` -- we need to retain the top-level `config` method for configuring user defaults (currently in design), so this stays.
* `list_packages` -- this method can target both the local registry and a remove `Bucket` registry, so it needs to live at the top level (for now)
* `delete_package` -- same as `list_packages`.

ResidentMario:
The diff coverage component of `codecov` is failing because this PR moved the untested `ls` function without implementing tests for it. `ls` is untested because its behavior is considered temporary and subject to change, although we never quite converged on _what_ the behavior should be.

api/python/t4/bucket.py
```diff
@@ -199,8 +200,47 @@ def delete(self, key):
         Raises:
             if delete fails
         """
+        if key[-1] == '/':
```
dimaryaz: Maybe raise an error if `key` is empty - e.g. `if not key or key.endswith('/'):`

ResidentMario: Good idea, but I think this is a separate error. I went ahead added a separate `if not key` check.

PR #193, opened 2018-12-20
Maintenance: remove nearly-empty requirements.txt

ResidentMario:
The sole contents of `requirements.txt` is `pylint=<some version>`. Not very useful.

A top level `requirements.txt` isn't very meaningful anyway, as this repository contains a dual `python` (SDK) and `javascript` (webapp) codebase. Requirements for either ought to be documented in their respective folders: `api/python` (which has `setup.py`) and `catalog` (which has `packages.json`).


akarve:
I take your points and am open to deleting this file, but I _think_ @stevededalus added it so that developers could run `pylint` anywhere in the repo? `pylint` is used by the code commit hooks. Maybe rename to `dev-requirements.txt`?

stevededalus:
I'm fine with removing it, as long as you verify/update the githooks to point to the right place to run.

ResidentMario:
@stevededalus I don't have admin on the repo so I can't be the one to do that verification. 😞

PR #196, opened 2018-12-21
Default local and remote registries

meffij:


api/python/t4/util.py
```diff
@@ -215,4 +223,12 @@ def __repr__(self):
 def validate_package_name(name):
     """ Verify that a package name is two alphanumerics strings separated by a slash."""
     if not re.match(PACKAGE_NAME_FORMAT, name):
-        raise QuiltException("Invalid package name, must contain exactly one /.")
\ No newline at end of file
+        raise QuiltException("Invalid package name, must contain exactly one /.")
+
+def load_config():
+    config_template = read_yaml(CONFIG_TEMPLATE)
```
dimaryaz: Do this in `else:`?

api/python/tests/integration/test_packages.py
```diff
@@ -69,6 +69,54 @@ def test_build(tmpdir):
...
 
+@patch('appdirs.user_data_dir', lambda x,y: os.path.join('test_appdir', x))
+def test_default_registry(tmpdir):
+    new_pkg = Package()
+
+    # Create a dummy file to add to the package.
+    test_file_name = 'bar'
+    with open(test_file_name, "w") as fd:
+        fd.write('test_file_content_string')
+        test_file = Path(fd.name)
```
dimaryaz: Btw, `Path` has a shortcut for writing files -`write_text(...)` and `write_binary(...)`. Much simpler than `with open(...)`.

api/python/t4/packages.py
```diff
@@ -318,7 +319,7 @@ def install(cls, name, registry, pkg_hash=None, dest=None, dest_registry=None):
             A new Package that points to files on your local machine.
         """
         if dest_registry is None:
-            dest_registry = BASE_PATH
+            dest_registry = load_config().get('default_local_registry', None)
```
dimaryaz: Probably should add a helper for this.

api/python/t4/packages.py
```diff
@@ -337,6 +338,10 @@ def browse(cls, name=None, registry=None, pkg_hash=None):
             registry(string): location of registry to load package from
             pkg_hash(string): top hash of package version to load
         """
+        if not registry:
```
dimaryaz: `if registry is None`, for consistency with the rest of the code.

api/python/t4/packages.py
```diff
@@ -657,6 +665,7 @@ def build(self, name=None, registry=None, message=None):
             manifest.getvalue(),
             registry_prefix + '/packages/' + hash_string
         )
+        print(registry_prefix + '/packages/' + hash_string)
```
dimaryaz: Remove it

api/python/t4/packages.py
```diff
@@ -832,8 +841,12 @@ def push(self, name, dest, registry=None, message=None):
         validate_package_name(name)
         self._set_commit_message(message)
 
-        if registry is None:
-            registry = dest
+        if not registry:
```
dimaryaz: Hm... are you intentionally changing it from `is None`?

api/python/t4/util.py
```diff
@@ -43,7 +43,15 @@
 # Used to satisfy search queries
 # elastic_search_url: https://example.com/es
 elastic_search_url:
-"""
+
+# default_local_registry: <url string, default: local appdirs>
+# default target registry for operations like install and build
+default_local_registry: {}
```
dimaryaz: Put it in quotes, just in case. I feel like URLs could cause problems in YAML... But they can't contain quotes, so this should be good enough.

PR #199, opened 2018-12-27
Catalog: guest (anonymous) access to buckets

nl0:
Allow guest access to bucket browsing / search (using configurable guest credentials) if the bucket allows it.

To be merged after #164

### DIFF
--- a/catalog/app/containers/AWSAuth/Provider.js
+++ b/catalog/app/containers/AWSAuth/Provider.js
@@ -58,12 +58,13 @@ export default RT.composeComponent('AWSAuth.Provider',
   }),
   ({ children, storage }) => (
     <Config.Inject>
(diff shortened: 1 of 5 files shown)


nl0:
rebased

meffij:
Should `guestCredentials` be an optional entry in the config? @akarve do we have a sane default for guest credentials? Will most normal deployments have guest credentials?

akarve:
We should skip guest credentials. Not only because it's questionable security, but also because we don't want to be trafficking key/secret pairs.

akarve:
Withdrawing my objection. Guest credentials OK for now.

meffij:
We do not intend to use this on any wide scale

We intend to switch to providing temporary credentials through a registry endpoint

Customers should not deploy this

PR #201, opened 2018-12-28
Implement a .quiltignore feature

ResidentMario:
Currently when you call `t4.Package.set_dir` you drop _all_ of the files in the current directory into the package. This is inconvenient because it will include files that you probably don't want to include in the package, like `.git`, `.DS_Store`, etcetera.

Also `t4` best practice is managing parts of the project folders you end up with using `t4` and parts of it using `git` et al. If we don't support a conventional `.gitignore` equivalent we make it a lot harder for users to snapshot and sync their projects quickly, they would have to run `set` file-by-file. This feature has been on my radar for a while now, and here's an implementation.

### DIFF
--- a/api/python/t4/util.py
+++ b/api/python/t4/util.py
@@ -247,6 +247,16 @@ def get_remote_registry():
 
 
 def quiltignore_filter(paths, ignore_rules, url_scheme):
+    """Given a list paths, filter out the paths which are captured by the given
+    ignore rules.
+
+    Args:
+        paths (list): a list or iterable of paths
+        ignore_rules (list): a list or iterable of ignore rules, in Unix shell
+            style wildcard format
+        url_scheme (str): the URL scheme, only the "file" scheme is currently
+            supported
+    """
     if url_scheme == 'file':
         from fnmatch import fnmatch
 
@@ -259,10 +269,14 @@ def quiltignore_filter(paths, ignore_rules, url_scheme):
 
         for ignore_rule in ignore_rules:
             ignore_rule = os.getcwd() + '/' + ignore_rule
-            files = set(n for n in files if not fnmatch(n, ignore_rule))
-            dirs = set(n for n in dirs if not fnmatch(n.as_posix() + "/", ignore_rule))
 
-        import pdb; pdb.set_trace()
+            for dir in dirs:
+                if fnmatch(dir.as_posix() + "/", ignore_rule):
+                    files = set(n for n in files if dir not in n.parents)
+                    dirs = dirs - {dir}
+
+            files = set(n for n in files if not fnmatch(n, ignore_rule))
+
         return files.union(dirs)
     else:
         raise NotImplementedError
--- a/api/python/t4/util.py
+++ b/api/python/t4/util.py
@@ -5,6 +5,7 @@ import json
 import os
 from urllib.parse import parse_qs, unquote, urlparse
 from urllib.request import url2pathname
+from fnmatch import fnmatch
 
 # backports
 from six.moves import urllib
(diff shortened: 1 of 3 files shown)


api/python/tests/integration/test_packages.py
```diff
@@ -347,6 +347,22 @@ def test_local_set_dir(tmpdir):
     # todo nested at set_dir site or relative to set_dir path.
     assert (bazdir / 'baz').resolve().as_uri() == pkg['my_keys/baz'].physical_keys[0]
 
+    # Verify ignoring files in the presence of a dot-quiltignore
+    with open('.quiltignore', 'w') as fd:
+        fd.write('foo\nbar')
```
stevededalus: tiny nit: can you do

```
fd.write('foo')
fd.write('bar')
```
less performant but more readable

api/python/t4/util.py
```diff
@@ -244,3 +244,42 @@ def get_local_registry():
...
+
+    Args:
+        paths (list): a list or iterable of paths
+        ignore_rules (list): a list or iterable of ignore rules, in Unix shell
+            style wildcard format
+        url_scheme (str): the URL scheme, only the "file" scheme is currently
+            supported
+    """
+    if url_scheme == 'file':
+        from fnmatch import fnmatch
```
stevededalus: Can you move this to the top of file unless there's a compelling reason not to? and if so comment

api/python/t4/util.py
```diff
@@ -244,3 +244,42 @@ def get_local_registry():
...
+        for ignore_rule in ignore_rules:
+            ignore_rule = os.getcwd() + '/' + ignore_rule
+
+            for dir in dirs:
+                # copy git behavior --- git matches paths and directories equivalently.
+                # e.g. both foo and foo/ will match the ignore rule "foo"
+                # but only foo/ will match the ignore rule "foo/"
+                if fnmatch(dir.as_posix() + "/", ignore_rule) or fnmatch(dir.as_posix(), ignore_rule):
+                    files = set(n for n in files if dir not in n.parents)
+                    dirs = dirs - {dir}
```
stevededalus: Does this work for multiple dirs? I think it might break the iterator:
```
>>> foo = range(10)
>>> for x in foo:
...     foo.remove(x)
>>> foo
[1, 3, 5, 7, 9]
```

api/python/tests/integration/test_packages.py
ResidentMario: Actually it should be:

```
fd.write('foo\n')
fd.write('bar')
```

`write` doesn't automatically line break between writes ([ref](https://stackoverflow.com/questions/2918362/writing-string-to-a-file-on-a-new-line-every-time)), which kind of makes sense.

api/python/t4/util.py
```diff
@@ -244,3 +245,50 @@ def get_local_registry():
...
+            else:
+                dirs.append(path)
+
+        for ignore_rule in ignore_rules:
+            ignore_rule = os.getcwd() + '/' + ignore_rule
+
+            if dirs:
+                dir_idx = 0
+                while True:
+                    dir = dirs[dir_idx]
```
stevededalus: dir is a built in function in python, might be good to avoid using as a var name.

ResidentMario: Done (and, OK, I'll hold to this convention in the future).

api/python/t4/util.py
```diff
@@ -244,3 +245,50 @@ def get_local_registry():
...
+
+                    # copy git behavior --- git matches paths and directories equivalently.
+                    # e.g. both foo and foo/ will match the ignore rule "foo"
+                    # but only foo/ will match the ignore rule "foo/"
+                    if fnmatch(dir.as_posix() + "/", ignore_rule) or fnmatch(dir.as_posix(), ignore_rule):
+                        files = set(n for n in files if dir not in n.parents)
+                        dirs = dirs[:dir_idx] + dirs[dir_idx + 1:]
+                    else:
```
stevededalus: i think it's easier to read with just a copy, i.e.

```
filtered_dirs = dirs[:]
for dir in dirs:
    if ...
        filtered_dirs.remove(dir)
...
files.union(filtered_dirs)
```

PR #204, opened 2019-01-07
Re-implement the upload/download/copy code.

dimaryaz:
We're no longer using the built-in S3 transfer manager, and instead implementing all the logic ourselves.
Changes:
- No more hacks to get the response metadata
- Upload/download/copy multiple sets of URLs in parallel
  - In particular, package push is parallelized
  - Local copies are also parallelized, but probably shouldn't be. TODO.
- Single progress bar for package push (well, two, actually: hashing and copying)
- Fixed versioned URLs fixed in package push

Some regressions:
- No more ETag optimization; will add it back eventually
- Progress callbacks are a bit coarse

### DIFF
--- a/api/python/tests/integration/data/t4_manifest.jsonl
+++ b/api/python/tests/integration/data/t4_manifest.jsonl
@@ -2,4 +2,3 @@
 {"logical_key":"foo","physical_keys":["s3://my_bucket/my_data_pkg/foo"],"hash":{"type":"SHA256","value":"abcd"},"size":1024,"meta":{}}
 {"logical_key":"bar.csv","physical_keys":["s3://my_bucket/my_data_pkg/bar.csv"],"hash":{"type":"SHA256","value":"abcd"},"size":1024,"meta":{}}
 {"logical_key":"baz/bat","physical_keys":["s3://my_bucket/my_data_pkg/baz/bat"],"hash":{"type":"SHA256","value":"abcd"},"size":1024,"meta":{}}
-{"logical_key":"my_data_pkg/data_frames/","physical_keys":["s3://my_bucket/my_data_pkg/data_frames/"],"hash":{"type":"SHA256","value":"abcd"},"size":1024,"meta":{}}
--- a/api/python/tests/integration/test_packages.py
+++ b/api/python/tests/integration/test_packages.py
@@ -6,7 +6,7 @@ import shutil
 from urllib.parse import urlparse
 
 import jsonlines
-from mock import patch, call
+from mock import patch, call, ANY
 import pytest
 
 import t4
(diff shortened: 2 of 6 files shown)


api/python/tests/integration/test_packages.py
```diff
@@ -249,53 +248,49 @@ def test_fetch(tmpdir):
...
-
-        # Get the second argument (destination) from the non-keyword args list
-        bytes_dest_args = [x[0][1] for x in bytes_mock.call_args_list]
-        file_dest_args = [x[0][1] for x in file_mock.call_args_list]
+        new_pkg.push('Quilt/package', 's3://my_test_bucket/')
 
         # Manifest copied
-        assert 's3://my_test_bucket/.quilt/packages/' + new_pkg.top_hash() in bytes_dest_args
-        assert 's3://my_test_bucket/.quilt/named_packages/Quilt/package_name/latest' in bytes_dest_args
+        print(bytes_mock.call_args_list)
```
stevededalus: stale print statement?

api/python/tests/integration/data/t4_manifest.jsonl
```diff
@@ -2,4 +2,3 @@
 {"logical_key":"foo","physical_keys":["s3://my_bucket/my_data_pkg/foo"],"hash":{"type":"SHA256","value":"abcd"},"size":1024,"meta":{}}
 {"logical_key":"bar.csv","physical_keys":["s3://my_bucket/my_data_pkg/bar.csv"],"hash":{"type":"SHA256","value":"abcd"},"size":1024,"meta":{}}
 {"logical_key":"baz/bat","physical_keys":["s3://my_bucket/my_data_pkg/baz/bat"],"hash":{"type":"SHA256","value":"abcd"},"size":1024,"meta":{}}
-{"logical_key":"my_data_pkg/data_frames/","physical_keys":["s3://my_bucket/my_data_pkg/data_frames/"],"hash":{"type":"SHA256","value":"abcd"},"size":1024,"meta":{}}
```
stevededalus: any reason there were deleted in this PR?

api/python/tests/integration/test_packages.py
dimaryaz: Oh yeah, will remove

api/python/tests/integration/data/t4_manifest.jsonl
dimaryaz: They end with `/` - but directories don't make sense as keys, right? The existing code somehow ignored that - but I've added sanity checks everywhere in the new code, so the tests started failing.

stevededalus: Ah, cool. We do allow directory level logical keys, but they should not have physical keys (they are used for storing directory level metadata)

PR #208, opened 2019-01-08
Use default parameters for code dependency hashes instead of Mapping

meffij:
Reorganize hashes for code dependencies.

This should help mitigate the current status quo where most stacks are on a non-standard template by allowing us to change the code dependencies deployed to a stack without editing a template.

TODO: investigate https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-grouping


meffij:
Dima pointed out that with default parameters, the hashes won't update after the initial deployment unless manually changed, which is not great. Closing this PR

PR #211, opened 2019-01-09
Disallow directories in package keys

dimaryaz:
Also clean up some version ID handling.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -22,7 +22,7 @@ from .exceptions import PackageException
 from .formats import FormatRegistry
 from .util import (
     QuiltException, fix_url, get_local_registry, get_remote_registry,
-    get_install_location, get_package_registry, parse_file_url, parse_s3_url, 
+    get_install_location, get_package_registry, make_s3_url, parse_file_url, parse_s3_url, 
     validate_package_name, quiltignore_filter
 )
 
@@ -576,9 +576,7 @@ class Package(object):
             for obj in objects:
                 if not obj['IsLatest']:
                     continue
-                obj_url = 's3://%s/%s' % (src_bucket, quote(obj['Key']))
-                if obj['VersionId'] != 'null':  # Yes, 'null'
-                    obj_url += '?versionId=%s' % quote(obj['VersionId'])
+                obj_url = make_s3_url(src_bucket, obj['Key'], obj.get('VersionId'))
                 entry = PackageEntry([obj_url], None, None, None)
                 logical_key = obj['Key'][len(src_key):]
                 # TODO: Warn if overwritting a logical key?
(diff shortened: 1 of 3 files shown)


api/python/t4/packages.py
```diff
@@ -749,16 +747,19 @@ def set(self, logical_key, entry, meta=None):
         Returns:
             self
         """
+        if not logical_key or logical_key.endswith('/'):
```
stevededalus: Ah, so we want to allow directory logical keys for attaching metadata to them, we just don't want them to have corresponding physical keys

api/python/t4/packages.py
```diff
@@ -749,16 +747,19 @@ def set(self, logical_key, entry, meta=None):
...
+            raise QuiltException("Invalid logical_key: %r; cannot be a directory" % logical_key)
```
stevededalus: still worth checking that it exists though

api/python/t4/packages.py
dimaryaz: Hm... So what do you pass as `entry` when creating a directory?

stevededalus: I assume None. @ResidentMario  ?

ResidentMario: @dimaryaz Let's make `entry` entirely optional. If left unspecified, try to `set` with the same `logical_key` and `entry` value (e.g. `.set('foo.csv')` == `.set('foo.csv', 'foo.csv')`).

This is something I've been thinking about that would reduce the amount of unnecessary typing when using `set` on regular old files as well. If we do this then we don't need to pass `None` explicitly when using `set` on a directory either. Kills two birds with one stone.

dimaryaz: Not sure this solves everything, though. What if the directory is not empty?

In any case, do we currently support creating empty directories? We're supposed to, but I can't find the code for that. Right now, I just want to make sure my change isn't breaking anything - and we can implement `None`, etc. as a follow-up.

ResidentMario: `set_dir` will let you set an empty directory. I don't have a comment on the rest of the changelist because I haven't read it in detail.

dimaryaz: Anyways... if something is broken, let me know, and I'll fix it.

stevededalus: I think you just need to allow the logical key to be None, don't do anything different than the current state. Only raise an exception if the logical key is a directory.

ResidentMario:
My two cents:

* `pkg.set('foo/', None)` should error out; you should be setting directories with `set_dir`. E.g. the "create an empty directory" behavior should be handled by `pkg.set_dir('foo/', None)`.
* `pkg.set('', None)` should error out, at least that's what I thought. Is empty string (`''`) significant? I thought we were using forward slash (`/`) to indicate the package root?

dimaryaz:
Ok, so currently, my code will error out on both `pkg.set('foo/', None)` and `pkg.set('', None)`.
I think you guys (@stevededalus and @ResidentMario) have different thoughts on how it should work - so we need to make a decision here.

ResidentMario:
I believe @stevededalus agrees with you that we should use `""` as a synonym for `"/"` because S3 uses `""` as its symbol for root? In that case, as long as the behavior of `""` and `"/"` is equivalent throughout the API, I'm satisfied.

dimaryaz:
Ok. What about `pkg.set('foo/', None)`? Does that do anything?

Also, the spec doesn't have any examples of how `set` should work...

PR #216, opened 2019-01-11
Implement CORS headers and JSON response type

dimaryaz:
- Return CORS headers if the origin is correct
- Return JSON rather than plain HTML if output=json is specified
- Add a parameter for nbconvert template file
- Add a parameter to explicitly set input type

### DIFF
--- a/lambdas/nbconvert/index.py
+++ b/lambdas/nbconvert/index.py
@@ -1,13 +1,12 @@
-from enum import Enum
 import json
 import os
-import urllib.request
+from tempfile import NamedTemporaryFile
 
 from jsonschema import Draft4Validator, ValidationError
 from nbconvert import HTMLExporter
 import nbformat
 import pandas as pd
-from urllib.parse import urlparse, unquote
+import requests
 
 WEB_ORIGIN = os.environ['WEB_ORIGIN']
 
@@ -18,17 +17,11 @@ SCHEMA = {
         'url': {
             'type': 'string'
         },
-        'output': {
-            'enum': ['html', 'json']
-        },
         'input': {
             'enum': ['ipynb', 'parquet']
-        },
-        'nbconvert_template': {
-            'enum': ['full', 'basic']
         }
     },
-    'required': ['url'],
+    'required': ['url', 'input'],
     'additionalProperties': False
 }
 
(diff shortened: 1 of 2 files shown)


lambdas/nbconvert/index.py
```diff
@@ -1,34 +1,105 @@
...
+
         response = urllib.request.urlopen(url).read().decode()
         notebook = nbformat.reads(response, 4)
-        body, resources = html_exporter.from_notebook_node(notebook)
+        result, _ = html_exporter.from_notebook_node(notebook)
+    elif input_type == 'parquet':
+        df = pd.read_parquet(url)
+        result = df._repr_html_()
+    else:
+        assert False
```
akarve: we should say something, like "Unexpected file extension". better yet, print what we found.

lambdas/nbconvert/index.py
```diff
@@ -1,34 +1,105 @@
...
-        body = df._repr_html_()
-    else:
+    url = params['url']
+    input_type = params.get('input')
+
+    # TODO(dima): Make it required.
+    if input_type is None:
+        parsed_url = urlparse(url)
+        path = unquote(parsed_url.path)
+        if path.endswith('.parquet'):
```
akarve: what guarantees that path is always lowercase?

lambdas/nbconvert/index.py
```diff
@@ -1,34 +1,105 @@
...
+            input_type = 'parquet'
+        else:
+            input_type = 'ipynb'
+
+
+    if input_type == 'ipynb':
+        html_exporter = HTMLExporter()
+        html_exporter.template_file = params.get('nbconvert_template', 'full')
+
         response = urllib.request.urlopen(url).read().decode()
```
akarve: can you write `urlopen`() to disk instead? will no longer use pandas to read URLs, can do more efficient I/O, won't need to rewind byte streams, use same disk blob for all extension types, etc.

lambdas/nbconvert/index.py
dimaryaz: Well, what guarantees that the extension is going to be `.parquet`, regardless of the case? What if the URL is something like `https://example.com/redirect?url=.../file.parquet`?

The point is, this can't work, so we should send the `input=...` parameter.

(Also, I didn't actually change this logic - just moved it around.)

lambdas/nbconvert/index.py
dimaryaz: Ok

lambdas/nbconvert/index.py
```diff
@@ -1,34 +1,105 @@
...
         notebook = nbformat.reads(response, 4)
-        body, resources = html_exporter.from_notebook_node(notebook)
+        result, _ = html_exporter.from_notebook_node(notebook)
```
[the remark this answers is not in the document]
dimaryaz: This is just a variable. Do you mean, just return JSON and drop plain HTML support?

lambdas/nbconvert/index.py
dimaryaz: This can't happen cause of the parameter validation at the top.

akarve: then we shouldn't have this `else`? idk if pylint asks for it then fine.

lambdas/nbconvert/index.py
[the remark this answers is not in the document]
akarve: i mean the response body should always be a JSON string with an `html` key. then, for every preview type, the client can *always* look in `body.html`, and optionally `body.meta`... instead of doing different things for each file type.

dimaryaz:
Ping?

PR #218, opened 2019-01-14
Filter returns a filtered package instead of a list of tuples

ResidentMario:
This PR tweaks the `t4.Package.filter` method to return a filtered copy of a package instead of a list of tuples. This makes more sense for the use case that `filter` was intended for (selecting just parts of packages).

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -983,10 +983,12 @@ class Package(object):
         if include_directories:
             for lk, _ in self._walk_dir_meta():
                 if not f(lk, self[lk.rstrip("/")]):
-                    excluded_dirs = excluded_dirs.union({lk})
+                    excluded_dirs.add(lk)
 
         for lk, entity in self.walk():
-            if f(lk, entity) and not any(p in excluded_dirs for p in pathlib.Path(lk).parents):
+            if (not any(p in excluded_dirs 
+                        for p in pathlib.PurePosixPath(lk).parents)
+                    and f(lk, entity)):
                 p.set(lk, entity)
 
         return p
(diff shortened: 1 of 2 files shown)


api/python/t4/packages.py
```diff
@@ -977,14 +977,19 @@ def filter(self, f, include_directories=False):
...
         """
+        p = Package()
+
+        excluded_dirs = set()
         if include_directories:
             for lk, _ in self._walk_dir_meta():
-                if f(lk, self[lk.rstrip("/")]):
-                    yield (lk, self[lk.rstrip("/")])
+                if not f(lk, self[lk.rstrip("/")]):
+                    excluded_dirs = excluded_dirs.union({lk})
```
dimaryaz: `excluded_dirs.add(lk)`

api/python/t4/packages.py
```diff
@@ -977,14 +977,19 @@ def filter(self, f, include_directories=False):
...
 
         for lk, entity in self.walk():
-            if f(lk, entity):
-                yield (lk, entity)
+            if f(lk, entity) and not any(p in excluded_dirs for p in pathlib.Path(lk).parents):
```
dimaryaz: I think you need a `PurePosixPath`, since these are logical keys

dimaryaz: Also... I think you should check the `excluded_dirs` first, then call `f`. It's probably reasonable for the user to assume `f` will never be called on keys in directories that got excluded - so it may raise an exception, etc.

api/python/t4/packages.py
ResidentMario: Done

api/python/t4/packages.py
ResidentMario: Great point! Done.

PR #221, opened 2019-01-17
Rename the "nbconvert" lambda to "preview"

dimaryaz:
It's now used for parquet previews, too, not just notebooks.


dimaryaz:
Yes - but the template refers to both a name and a specific git revision, so I'll update it when I merge this and it's ready to deploy

stevededalus:
Sgtm

PR #227, opened 2019-01-25
T4 private cloud API additions

meffij:
Requires a Quilt 2 registry to provide endpoints


meffij:
WIP updated for new endpoints

api/python/t4/api.py
```diff
@@ -449,3 +450,120 @@ def config(*autoconfig_url, **config_values):
         write_yaml(local_config, CONFIG_PATH, keep_backup=True)
 
     return HeliumConfig(CONFIG_PATH, local_config)
+
+def create_role(name, arn=None):
```
stevededalus: I'm not really sure how much utility these methods add as written, they just wrap the single line `get_session()._REST_CALL_` . It seems like the response should be translating the response to a python object at least. And possibly operating on a python object in the case of edit (so`role_foo.edit_role(name,arn)`)

api/python/t4/__init__.py
```diff
@@ -12,7 +12,8 @@
     list_packages,
     search,
     config,
-    delete_package
+    delete_package,
+    login_user_pass
```
stevededalus: does this need to import all the `role` endpoints you just added

api/python/t4/api.py
```diff
@@ -449,3 +450,120 @@ def config(*autoconfig_url, **config_values):
...
+    }
+    response = session.post(
+        "{url}/api/users/set_role".format(
+            url=get_registry_url()
+        ),
+        data=json.dumps(data)
+    )
+    return response
+
+def login_user_pass(username, password):
```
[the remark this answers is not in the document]
stevededalus: Let's make pw an environment variable until we can update catalogue to handle the login/redirect stuff

api/python/tests/test_api.py
```diff
@@ -87,3 +88,42 @@ def test_put_copy_get(self):
...
+        mock_response = Mock()
+        mock_response.ok = True
+        mock_session.post.return_value = mock_response
+
+        mock_creds_response = Mock()
+        creds_data = {
+            'AccessKeyId': 'asdf',
+            'SecretAccessKey': 'asdf',
+            'SessionToken': 'asdf',
+            'Expiration': datetime.utcnow().isoformat() + '-02:00' # so it's not expired yet
```
stevededalus: I think this is futzing the timezone, it's probably clearer to do something like
```
+ timedelta(hours=2)
```
then format it to an iso string

api/python/t4/__init__.py
meffij: they hang off `t4.admin` now

api/python/t4/data_transfer.py
```diff
@@ -48,6 +50,18 @@
 s3_threads = 4
 
 
+def _update_credentials():
+    # TODO: use new credentials object for clients
+    assert get_credentials(), "This method should not be called unless CREDENTIALS is populated"
```
stevededalus: No asserts in non-test code.

api/python/t4/api.py
```diff
@@ -449,3 +450,120 @@ def config(*autoconfig_url, **config_values):
...
+    data['arn'] = new_arn or old_data['arn']
+
+    response = session.put(
+        "{url}/api/roles/{role_id}".format(
+            url=get_registry_url(),
+            role_id=role_id
+            ),
+            data=data
+        )
+    assert response.ok
```
stevededalus: No asserts in non-test code.

api/python/t4/session.py
```diff
@@ -176,6 +176,24 @@ def login():
...
+        'username': username,
+        'password': password
+    }
+    response = requests.post(
+        "{url}/api/login".format(
+            url=get_registry_url()
+        ),
+        json=data
+    )
+    assert response.ok
```
stevededalus: No asserts in non-test code.

api/python/t4/api.py
```diff
@@ -449,3 +450,121 @@ def config(*autoconfig_url, **config_values):
...
+        "{url}/api/roles".format(
+            url=get_registry_url()
+            ),
+            data=json.dumps({
+                'name': name,
+                'arn': arn
+            })
+        )
+    if not response.ok:
+        message = response.json()['message']
```
dimaryaz: You could also get a 500, which might not be valid JSON. Also, `session.post` could raise an exception for connection errors, etc.

We probably want to take the error handling from Quilt2. Though could punt on that for now, I guess

api/python/t4/api.py
```diff
@@ -449,3 +450,121 @@ def config(*autoconfig_url, **config_values):
...
+    data = {}
+    data['name'] = new_name or old_data['name']
+    data['arn'] = new_arn or old_data['arn']
+
+    response = session.put(
+        "{url}/api/roles/{role_id}".format(
+            url=get_registry_url(),
+            role_id=role_id
+            ),
+            data=data
```
dimaryaz: Hm... No `json.dumps`?

Though actually, you should be able to use `json=` instead of `data=`

api/python/t4/session.py
```diff
@@ -190,6 +190,10 @@ def login_with_token(refresh_token):
 
     clear_session()
 
+    # use registry-provided credentials if available
+    if try_registry_credentials():
```
dimaryaz: This seems like a bad idea - I think we should make it explicit. Also, this could fail due to things like network errors.

api/python/t4/api.py
```diff
@@ -449,3 +450,121 @@ def config(*autoconfig_url, **config_values):
...
+
+def list_roles():
+    """
+    List configured roles. Admins only.
+    """
+    session = get_session()
+    response = session.get(
+        "{url}/api/roles".format(
+            url=get_registry_url()
+        ))
```
stevededalus: Any reason why you dropped     
```
if not response.ok:
        raise QuiltException("...")
```

here and don't have it inset_role?

api/python/t4/session.py
```diff
@@ -201,3 +205,46 @@ def logout():
...
+def try_registry_credentials():
+    session = get_session()
+    try:
+        response = session.get(
+            "{url}/api/auth/get_credentials".format(
+                url=get_registry_url()
+            )
+        )
+        return True
+    except QuiltException:
```
dimaryaz: Can this even happen? You're using the requests session directly, right?

api/python/t4/api.py
meffij: the session raises an exception on its own if the response has a 4xx or 5xx code

api/python/t4/api.py
meffij: My intention is to use the error handling from Quilt2. I added this block by mistake, removing it now.

api/python/t4/api.py
```diff
@@ -449,3 +450,112 @@ def config(*autoconfig_url, **config_values):
...
+    Required Parameters:
+        name(string): name of role to create
+        arn(string): ARN of IAM role to associate with the Quilt role you are creating
+    """
+    session = get_session()
+    response = session.post(
+        "{url}/api/roles".format(
+            url=get_registry_url()
+            ),
+            data=json.dumps({
```
dimaryaz: Might as well use `json=` here, too?

api/python/t4/api.py
```diff
@@ -449,3 +450,112 @@ def config(*autoconfig_url, **config_values):
...
+    Required parameters:
+        role_id(string): ID of role you want to delete.
+    """
+    session = get_session()
+    response = session.delete(
+        "{url}/api/roles/{role_id}".format(
+            url=get_registry_url(),
+            role_id=role_id
+            )
+        )
```
dimaryaz: Return the response?

api/python/t4/api.py
```diff
@@ -449,3 +450,112 @@ def config(*autoconfig_url, **config_values):
...
+    session = get_session()
+    data = {
+        'username': username,
+        'role': role_name
+    }
+    response = session.post(
+        "{url}/api/users/set_role".format(
+            url=get_registry_url()
+        ),
+        data=json.dumps(data)
```
dimaryaz: `json=`

api/python/t4/api.py
```diff
@@ -449,3 +450,112 @@ def config(*autoconfig_url, **config_values):
...
+    """
+    List configured roles. Admins only.
+    """
+    session = get_session()
+    response = session.get(
+        "{url}/api/roles".format(
+            url=get_registry_url()
+        ))
+
+    return response.json()['results']
```
dimaryaz: Kind of inconsistent with the others

api/python/t4/api.py
```diff
@@ -449,3 +450,112 @@ def config(*autoconfig_url, **config_values):
...
+    data = {
+        'username': username,
+        'role': role_name
+    }
+    response = session.post(
+        "{url}/api/users/set_role".format(
+            url=get_registry_url()
+        ),
+        data=json.dumps(data)
+    )
```
dimaryaz: Return the response

api/python/t4/session.py
dimaryaz: But it's still here?

api/python/t4/api.py
stevededalus: Cool, thanks for clarifying

api/python/t4/api.py
stevededalus: If that doesn't make it into this PR, can you throw a TODO in?

meffij:
TODO pylint

PR #229, opened 2019-01-25
Catalog: search hotfix

nl0:
Catalog displayed "search not available" when the bucket config is present, but bucket name not included in `suggestedBuckets`.


nl0:
@akarve rtm

PR #234, opened 2019-01-29
Improve search docstring

meffij:


api/python/t4/bucket.py
```diff
@@ -78,7 +85,7 @@ def search(self, query, size=None):
         """
         if not self._search_endpoint:
             self.config()
-        return search(query, self._search_endpoint, size=size)
+        return search(query, self._search_endpoint, size=limit)
```
stevededalus: This is confusing, the doc string for that method says it returns objects with:
```
        size: size of object in bytes
```
Is the payload size limiting the number of object or sizes of returned objects? Can you update that docstring/possibly rename the function argument to be consistent.

meffij: The argument called `limit` is the maximum number of returned objects. I was calling it `size` before because that's what it's called in the ES API. I changed it because Aneesh correctly pointed out that it's confusing when `size` is also part of the returned objects. Pushing a commit now that further renames things to `limit`

PR #236, opened 2019-01-30
Catalog: display summary in package tree

nl0:
Display summary in package tree similar to regular bucket tree.

Limitations:

* objects linked in `quilt_summarize.json` are previewed and linked without version (not using the version from the package revision currently browsed, but the latest one from the bucket)
* object title links to that object's physical location (in bucket tree, not in package tree)
* thumbnails link to images' physical locations (in bucket tree, not in package tree)

Should be merged after #232 and #233


nl0:
@akarve rebased

PR #237, opened 2019-02-02
Reorganize versions menu, add relative date

akarve:


nl0:
@akarve i'd put all the secondary text on one line, so it's a little more wide rather than tall, but otherwise looks good

akarve:
@nl0 Good idea to put secondary on one line; I looked into this and, I guess because of overflow, it's laying out one two lines. I welcome an improvement in a future rev. I also thought about using `title=` for the version but that seemed to take away the nerd candy :)

PR #239, opened 2019-02-02
Bucket/requests: objectVersions: filter object by key

nl0:
Version list returned by S3 is only filtered by prefix, not key, so we need to filter it by key manually to drop irrelevant versions


nl0:
@akarve this one is also super simple, so i'm going to merge it as well

PR #245, opened 2019-02-07
Update registry defaulting behavior

ResidentMario:
A recent PR added a concept of a default registry to user configuration, allowing users to under-specify the parameters in `push` and use the API in a more convenient way.

This PR tweaks how `push` interacts with the default remote registry, using it to parameterize the `dest` kwarg instead of the `registry` kwarg. This is how I expected the API to work, and it makes more sense to me.

Old behavior:

```python
# assume that the default registry is 's3://my-bucket'
p = t4.Package()

# == p.push('foo/bar', dest='s3://other-bucket/foo/bar', registry='s3://my-bucket/')
p.push("foo/bar", "s3://other-bucket")
```

This is confusing because it looks like I'm pushing `foo/bar` to the `other-bucket` S3 bucket. However, instead I am pushing the _files_ to `s3://other-bucket`, and the _package manifest_ to `s3://my-bucket`. From my perspective as an end user when I go to visit `s3://other-bucket` it looks like my package disappeared!

New behavior:

```python
# assume that the default registry is 's3://my-bucket'
p = t4.Package()

# == p.push('foo/bar', dest='s3://my-bucket/foo/bar', registry='s3://my-bucket/')
p.push("foo/bar")   

# == p.push('foo/bar', dest='s3://other-bucket/foo/bar', registry='s3://other-bucket/')
p.push("foo/bar", dest="s3://other-bucket/foo/bar")

# raises; 'dest' is still a required argument
p.push("foo/bar", registry="s3://my-bucket/")
```

Changes:
* The default registry is now only used if only the package name is specified (e.g. `p.push('foo/bar')`).
* `registry` is inferred from `dest`. If the target is `s3://` the root of the bucket is passed as the registry, if the target is `file:///` the path is used as the registry.

If you want to deposit the files in one bucket and the manifest in another you can still do that with a fully specified `push`.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -859,34 +859,36 @@ class Package(object):
         validate_package_name(name)
         self._set_commit_message(message)
 
-        # if only a package name is set, set registry and dest to the default registry
-        if dest is None and registry is None:
-            registry = get_remote_registry()
-            if not registry:
-                raise QuiltException("No registry specified and no default remote "
-                                     "registry configured. Please specify a registry "
-                                     "or configure a default remote registry with t4.config")
+        if registry is None:
+            if dest is None:
+                # Only a package name is set, so set registry and dest to the default 
+                # registry.
+                registry = get_remote_registry()
+                if not registry:
+                    raise QuiltException("No registry specified and no default remote "
+                                        "registry configured. Please specify a "
+                                        "registry or configure a default remote "
+                                        "registry with t4.config")
 
-            dest = registry
-
-        # if dest is specified and registry is not use dest as the registry
-        elif dest is not None and registry is None:
-            parsed = urlparse(fix_url(dest))
-
-            # TODO: this should be a helper method?
-            if parsed.scheme == 's3':
-                bucket, _, _ = parse_s3_url(parsed)
-                registry = 's3://' + bucket
-            elif parsed.scheme == 'file':
-                registry = parsed.path
+                dest = registry
             else:
-                raise NotImplementedError
+                # The dest is specified and registry is not. Get registry from dest.
+                parsed = urlparse(fix_url(dest))
+                if parsed.scheme == 's3':
+                    bucket, _, _ = parse_s3_url(parsed)
+                    registry = 's3://' + bucket
+                elif parsed.scheme == 'file':
+                    registry = parsed.path
+                else:
+                    raise NotImplementedError
 
-        # specifying registry but not dest doesn't make sense
-        elif dest is None and registry is not None:
-            raise NotImplementedError
-
-        # if both dest and registry are specified no further work is needed
+        else:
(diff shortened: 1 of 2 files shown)


api/python/t4/packages.py
```diff
@@ -859,13 +859,35 @@ def push(self, name, dest, registry=None, message=None):
         validate_package_name(name)
         self._set_commit_message(message)
 
-        if registry is None:
+        # if only a package name is set, set registry and dest to the default registry
```
stevededalus: Make these comments sentences (Capital letter + punctuation)

api/python/t4/packages.py
```diff
@@ -859,13 +859,35 @@ def push(self, name, dest, registry=None, message=None):
...
+            if parsed.scheme == 's3':
+                bucket, _, _ = parse_s3_url(parsed)
+                registry = 's3://' + bucket
+            elif parsed.scheme == 'file':
+                registry = parsed.path
+            else:
+                raise NotImplementedError
+
+        # specifying registry but not dest doesn't make sense
+        elif dest is None and registry is not None:
```
stevededalus: this feels like a case where two nested if statements are more readable 

```
if dest is none:
    if registry is none:
...
```

api/python/tests/integration/test_packages.py
```diff
@@ -111,12 +111,13 @@ def test_default_registry(tmpdir):
...
             assert test_file.resolve().as_uri() == pkg['bar'].physical_keys[0]
 
+    # TODO: fix this test?
     with patch('t4.packages.get_remote_registry') as mock_config:
         mock_config.return_value = new_base_path
         new_pkg.push("Quilt/Test", Path(tmpdir, 'test_dest').resolve().as_uri())
         with open(out_path) as fd:
             pkg = Package.load(fd)
-            assert pkg['bar'].physical_keys[0].endswith('test_dest/Quilt/Test/bar')
+            assert pkg['bar'].physical_keys[0].endswith('test_default_registry0/bar')
```
stevededalus: what's going on with this `default_registry0` string?

api/python/tests/integration/test_packages.py
```diff
@@ -260,6 +261,7 @@ def test_load_into_t4(tmpdir):
 
         # Manifest copied
         top_hash = new_pkg.top_hash()
+        # FIXME: understand what is happening with these byte calls
```
stevededalus: What's this comment mean?

ResidentMario: Old debug line, removed.

api/python/tests/integration/test_packages.py
ResidentMario: Updated the test, it now actually exercises the `get_remote_registry` code path.

PR #249, opened 2019-02-08
Bucket/File: fix version date display

nl0:


nl0:
@akarve i'm gonna merge this trivial fix

PR #252, opened 2019-02-09
Add proper unit tests for preview lambda function

akarve:
And fix a few bugs. In a future PR we will run these unit tests as part of CI.


lambdas/preview/index.py
```diff
@@ -33,15 +31,14 @@
...
-
 def lambda_handler(event, _):
     """
     dynamically handle preview requests for bytes in S3
 
     caller must specify input_type (since there may be no file extension)
     """
-    params = event['queryStringParameters'] or {}
-    headers = event['headers'] or {}
+    params = event.get('queryStringParameters', {})
```
dimaryaz: Uhh that was intentional - that's how AWS gives you the data. If there are no query params, you get a `None`. It's stupid... But, need to adjust the unit test to simulate AWS, not adjust the code to fit the unit test :-P

akarve: Oh, OK, I'll revert that

lambdas/preview/index.py
```diff
@@ -9,12 +9,15 @@
...
-import pandas as pd
-import pyarrow as pa
+import pyarrow.parquet as pq
 import requests
 
-WEB_ORIGIN = os.environ['WEB_ORIGIN']
 
+ALLOWED_ORIGINS = [
+    'http://localhost:3000',
+    'http://localhost:3000/',
```
dimaryaz: I looked up the spec; there should be no trailing slash, so let's remove this one. (Though we could add a sanity check for `os.environ.get('WEB_ORIGIN')`, to make sure it doesn't have one.)

lambdas/preview/index.py
```diff
@@ -60,32 +64,54 @@ def lambda_handler(event, _):
...
+                info['num_row_groups'] = meta.num_row_groups
+                info['schema'] = {
+                    meta.schema.names[i]: {
+                        'logical_type': meta.schema.column(i).logical_type,
+                        'max_definition_level': meta.schema.column(i).max_definition_level,
+                        'max_repetition_level': meta.schema.column(i).max_repetition_level,
+                        'path': meta.schema.column(i).path,
+                        'physical_type': meta.schema.column(i).physical_type,
+                    }
+                    for i in range(len(meta.schema.names))
```
dimaryaz: `for i, name in enumerate(meta.schema.names)`

lambdas/preview/test/test_index.py
```diff
@@ -0,0 +1,79 @@
...
+from unittest.mock import patch
+import responses
+
+from ..index import lambda_handler
+
+MOCK_ORIGIN = 'https://mock.quiltdata.com'
+
+# pylint: disable=no-member
+class TestIndex():
+    """Class kto test various inputs to the main indexing function"""
```
dimaryaz: to

lambdas/preview/test/test_index.py
```diff
@@ -0,0 +1,79 @@
...
+        basedir = os.path.join(parent, 'data')
+        notebook = os.path.join(basedir, 'nb_1200727.ipynb')
+        with open(notebook, 'rb') as file_:
+            responses.add(
+                responses.GET,
+                self.FILE_URL,
+                body=file_.read(),
+                status=200)
+            # make it an ipynb event
+            self.S3_EVENT['queryStringParameters']['input'] = 'ipynb'
```
dimaryaz: Don't modify globals; this can affect other tests

lambdas/preview/test/test_index.py
```diff
@@ -0,0 +1,79 @@
...
+                self.FILE_URL,
+                body=file_.read(),
+                status=200)
+            # make it an ipynb event
+            self.S3_EVENT['queryStringParameters']['input'] = 'parquet'
+            resp = lambda_handler(self.S3_EVENT, None)
+            assert resp['statusCode'] == 200, f"Expected 200, got {resp['statusCode']}"
+            body = json.loads(resp['body'])
+            with open(info_response, 'r') as info_json:
+                expected = json.loads(info_json.read())
```
dimaryaz: `json.load(info_json)`

lambdas/preview/index.py
```diff
@@ -60,24 +59,24 @@ def lambda_handler(event, _):
 
     resp = requests.get(url)
     if resp.ok:
-        with NamedTemporaryFile() as fd:
```
dimaryaz: I kinda like `fd`, TBH. The trailing `_`s are weird. (If `pylint` is complaining about it, I can fix it.)

akarve: Yeah, `pylint`. I find this linting rule a tad silly but, at the same time, `fd` is myster meat. `f` makes more sense to me but not religious.

dimaryaz: Well, `fd` just means "file descriptor". (I guess it's technically a file object here, not a file descriptor... need to find out what the convention is.)

stevededalus: I think you can modify how permissive pylint is about different variable names if you want to allow things like fd: 
http://pylint.pycqa.org/en/1.9/user_guide/options.html#cmdoption-variable-rgx

PR #255, opened 2019-02-14
Provide default install location

ResidentMario:
Currently `t4.Package.install` must be parameterized with a `dest`, otherwise it doesn't know where to send the data, since we don't provide a default local location.

This PR adds two things:
* A default `install` `dest`. This will be `<appdirs>/.quilt/data/`; a neighbor of `<appdirs>/.quilt/named_packages` and `<appdirs>/.quilt/packages`.
* A new `default_install_location` flag that can be set using `t4.config`. If unset, the default install location described above will be used. If set, the file path this flag is set to will be used instead.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -21,8 +21,9 @@ from .data_transfer import (
 from .exceptions import PackageException
 from .formats import FormatRegistry
 from .util import (
-    QuiltException, BASE_PATH, fix_url, get_local_registry, get_remote_registry,
-    parse_file_url, parse_s3_url, validate_package_name, quiltignore_filter
+    QuiltException, fix_url, get_local_registry, get_remote_registry,
+    get_install_location, get_package_registry, parse_file_url, parse_s3_url, 
+    validate_package_name, quiltignore_filter
 )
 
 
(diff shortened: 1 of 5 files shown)


api/python/t4/packages.py
```diff
@@ -328,7 +323,7 @@ def install(cls, name, registry=None, pkg_hash=None, dest=None, dest_registry=No
         """
         if registry is None:
             registry = get_remote_registry()
-            if not registry:
+            if registry is None:
```
stevededalus: why the switch from `not` to `is none`?

ResidentMario: Well, `if not registry` will fire for anything that evaluates to falsey. `registry is None` will only fire if `registry` is truly `None`. Since we use `None` as our sentinel value (e.g. it's the default value when the user doesn't specify any other value), this is a more semantically correct way of performing a nullity check.

This matters if the user specifies e.g. the following:

```
def install(..., registry="")
```

While this isn't a _correct_ thing to do, it's more correct error behavior to return an error stating that '`""` is a not a valid registry argument' then to discard/ignore this argument and proceed to running the default-registry-setting code.

PR #257, opened 2019-02-14
Implement proper error handling

dimaryaz:
- Catch exceptions and return a 500
- Return CORS headers for every response, including errors


lambdas/preview/test/test_index.py
```diff
@@ -27,10 +28,42 @@ def _make_event(cls, query, headers=None):
...
+        notebook = BASE_DIR / 'nb_1200727.ipynb'
+        responses.add(
+            responses.GET,
+            self.FILE_URL,
+            body=notebook.read_bytes(),
+            status=200)
+        event = self._make_event({'url': self.FILE_URL, 'input': 'ipynb'}, {'origin': MOCK_ORIGIN})
+        with patch('nbconvert.HTMLExporter.from_notebook_node') as func:
+            # Simulate a typical nbconvert failure.
+            func.side_effect = TypeError("I didn't test my code, so it crashed.")
```
akarve: but i did test it

lambdas/preview/requirements.txt
```diff
@@ -0,0 +1,47 @@
+# TODO: add "../shared/" or "t4-lambda-shared" in a way that's not totally broken
```
akarve: what's the broken part? does requirements.txt not support relative paths?

lambdas/preview/requirements.txt
```diff
@@ -0,0 +1,47 @@
...
+parso==0.3.4
+pexpect==4.6.0
+pickleshare==0.7.5
+pluggy==0.8.1
+prompt-toolkit==2.0.8
+ptyprocess==0.6.0
+py==1.7.0
+pyarrow==0.12.0
+Pygments==2.3.1
+pytest==4.2.1
```
akarve: the [tests] extras are also in requirements.txt (pinned to a specific verision) -- but why would we send the tests requirements to lambda?

lambdas/shared/setup.py
```diff
@@ -0,0 +1,17 @@
+from setuptools import setup
```
akarve: this file could use a comment explaining intended use and that /lambdas/whatever will expect that `pip install .` has been called

lambdas/preview/index.py
```diff
@@ -32,28 +32,14 @@
     'additionalProperties': False
 }
 
-Draft4Validator.check_schema(SCHEMA)
-VALIDATOR = Draft4Validator(SCHEMA)
-
-def lambda_handler(event, _):
+@api(cors_origins=ALLOWED_ORIGINS)
+@validate(SCHEMA)
```
akarve: this should really be a JSON file; and `test_validator` should use the same file, instead of a made-up `dict`

lambdas/shared/t4_lambda_shared/decorator.py
```diff
@@ -0,0 +1,58 @@
...
+                body = 'Internal Server Error'
+                response_headers = {
+                    'Content-Type': 'text/plain'
+                }
+
+            origin = headers.get('origin')
+            if origin is not None and origin in cors_origins:
+                response_headers.update({
+                    'access-control-allow-origin': '*',
+                    'access-control-allow-methods': 'GET,POST',
```
akarve: might need HEAD at some point... I guess preview in doesn't need HEAD, so maybe @api gets a default kwarg `allow_methods=['GET', 'POST']`. i'm not religious about this but, at least for interacting with S3, we keep needing more and more methods than we expected.

lambdas/shared/tests/test_decorator.py
```diff
@@ -0,0 +1,81 @@
...
+Decorator tests
+"""
+
+from unittest import TestCase
+
+from t4_lambda_shared.decorator import api, validate
+
+class TestDecorator(TestCase):
+    def test_api_basic(self):
+        @api(cors_origins=['https://example.com'])
```
akarve: it would be nice to have a unit test that tosses a few broad errors/exceptions

lambdas/shared/t4_lambda_shared/decorator.py
```diff
@@ -0,0 +1,58 @@
...
+
+def api(cors_origins=[]):
+    def innerdec(f):
+        @wraps(f)
+        def wrapper(event, _):
+            params = event['queryStringParameters'] or {}
+            headers = event['headers'] or {}
+            try:
+                status, body, response_headers = f(params, headers)
+            except Exception:
```
akarve: this deserves a docstring mention: we will literally catch anything, e.g. KeyboardInterrupt (which shouldn't be a factor in lambda functions)

lambdas/preview/test/test_index.py
```diff
@@ -27,10 +28,42 @@ def _make_event(cls, query, headers=None):
 
     def test_bad(self):
         """send a known bad event (no input query parameter)"""
-        event = self._make_event({'url': self.FILE_URL})
+        event = self._make_event({'url': self.FILE_URL}, {'origin': MOCK_ORIGIN})
         resp = lambda_handler(event, None)
         assert resp['statusCode'] == 400, "Expected 400 on event without 'input' query param"
         assert resp['body'], 'Expected explanation for 400'
```
akarve: maybe something dumb about my test setup by pylint vomits all over this file with
```
test_index.py:33:0: E1126: Sequence index is not an int, slice, or instance with __index__ (invalid-sequence-index)
```
probably a bug we should suppress this message?

lambdas/shared/tests/test_decorator.py
```diff
@@ -0,0 +1,81 @@
+"""
```
akarve: worth a quick `pylint` on this file. i'm not huge on docstrings for tests but might as well set the bar :P

lambdas/preview/requirements.txt
dimaryaz: Too much to fit into one comment :)

It supports relative paths, but only relative to the current working directory, not to the current file. If I use `-r ../shared/requirements.txt`, that will work relative to the current file - but not `../shared/`. Using `file:../shared/` should in theory work - but doesn't.

If I were to say `t4-lambda-shared==0.0.1`, that would check that the package is installed, though wouldn't be able to install it if it weren't. That would be fine... Except if I use `pip install -t dir ...`, as I do when building the zip file - in which case, even if the package is already installed, it still tries to install it again.

So I've given up on fighting it for now.

lambdas/preview/requirements.txt
dimaryaz: Oh, good point

lambdas/shared/setup.py
dimaryaz: Eh, that's like literally what `setup.py` is for :)

lambdas/preview/index.py
dimaryaz: So pass a file path instead?

lambdas/shared/tests/test_decorator.py
dimaryaz: I'm already testing a `TypeError`, which I think is pretty broad?

lambdas/shared/t4_lambda_shared/decorator.py
dimaryaz: Nope - `KeyboardInterrupt` is not an `Exception`

lambdas/preview/requirements.txt
akarve: k, a link to this comment would be nice `https://github.com/quiltdata/t4/pull/257#discussion_r259040374`

lambdas/preview/index.py
akarve: yeah

lambdas/shared/setup.py
akarve: sorry, my logic was backwards. it's not about this directory but about lambdas/not_shared. I think a lambdas/README.md is worthwhile to mention this, but not religious. people will work in isolated folders and it's not super obvious that an external setup file is needed.

lambdas/shared/tests/test_decorator.py
akarve: sure; but that's in test_index right now. belongs in test_decorator, right? e.g. if we lift this code to another project there will be no unit test for errors in test_decorator.

dimaryaz: Ohh, you're right. I'll move the test here - it's not really testing the preview endpoint

lambdas/preview/index.py
dimaryaz: So I moved things around, and not sure this still makes sense. The validator tests are now in shared, but so the preview schema doesn't make sense there. The preview schema does get tested in `test_bad`; if that's not enough, I can add more things there.

akarve: this is OK for now. but, and i only just realized this, schema's should be stand-alone files so that they can be checked into catalog, verified by javascript, etc.

PR #265, opened 2019-02-15
Make Parquet table styling more like pandas

akarve:


nl0:
@akarve i've moved the styles

PR #266, opened 2019-02-15
Try to fix a test

dimaryaz:
Looks like files might not be sorted.


ResidentMario:
This was separately addressed in a now-merged PR from @meffij.

PR #268, opened 2019-02-16
Implement package module import

ResidentMario:
This PR adds the ability to import a package as a module directly inside of a Python file.

Quilt 2 serves as prior art for this. This PR uses a slightly cleaned-up Python 3.6 way of doing things.

### DIFF
--- a/api/python/t4/imports.py
+++ b/api/python/t4/imports.py
@@ -72,7 +72,6 @@ class DataPackageFinder:
             return None
         else:
             return ModuleSpec(fullname, DataPackageLoader())
-        return ModuleSpec(fullname, DataPackageFinder()) if 't4' in fullname else None
 
 
 def start_data_package_loader():
--- a/api/python/t4/imports.py
+++ b/api/python/t4/imports.py
@@ -15,7 +15,7 @@ class DataPackageLoader:
     """
 
     @classmethod
-    def create_module(cls, spec):
+    def create_module(cls, spec):  # pylint: disable=unused-argument
         """
         Module creator. Returning None causes Python to use the default module creator.
         """
@@ -54,6 +54,7 @@ class DataPackageLoader:
             raise NotImplementedError
 
 
+# pylint: disable=too-few-public-methods
 class DataPackageFinder:
     """
     Data package module loader finder. This class sits on `sys.meta_path` and returns the
@@ -61,7 +62,7 @@ class DataPackageFinder:
     """
 
     @classmethod
-    def find_spec(cls, fullname, path=None, target=None):
+    def find_spec(cls, fullname, path=None, target=None):  # pylint: disable=unused-argument
         """
         This functions is what gets executed by the loader.
         """
--- a/api/python/t4/imports.py
+++ b/api/python/t4/imports.py
@@ -49,6 +49,9 @@ class DataPackageImporter:
             module.__path__ = MODULE_PATH
             return module
 
+        else:
+            assert False
+
 
 # pylint: disable=too-few-public-methods
 class DataPackageFinder:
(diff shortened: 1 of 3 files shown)


ResidentMario:
CI failure is due to a flaky test addressed in PR#266.

api/python/t4/imports.py
```diff
@@ -0,0 +1,81 @@
...
+                    module.__dict__.update({pkg_name: Package.browse(pkg)})
+
+            module.__path__ = MODULE_PATH
+            return module
+
+        else:
+            # an implementation for subpackage imports exists, but this has significant
+            # consistency issues. For now let's avoid, but you can see the full code at
+            # https://github.com/ResidentMario/package-autorelaod/blob/master/loader.py
+            raise NotImplementedError
```
dimaryaz: This seems wrong - don't think imports are allowed to raise random exceptions. Maybe do this check in `DataPackageFinder` and return `None`?

api/python/t4/imports.py
```diff
@@ -0,0 +1,81 @@
...
+            return module
+
+        elif len(name_parts) == 3:  # e.g. module.__name__ == t4.data.foo
+            namespace = name_parts[2]
+
+            # we do not know the name the user will ask for, so populate all valid names
+            for pkg in list_packages():
+                pkg_user, pkg_name = pkg.split('/')
+                if pkg_user == namespace:
+                    module.__dict__.update({pkg_name: Package.browse(pkg)})
```
dimaryaz: Just use `module.__dict__[pkg_name] = ...`?

api/python/t4/imports.py
```diff
@@ -0,0 +1,81 @@
+"""Implementation of the Python Quilt T4 data package loader."""
+
+from importlib.machinery import ModuleSpec
+import sys
+
+from t4.util import BASE_PATH
+from t4 import list_packages, Package
+
+
+MODULE_PATH = [BASE_PATH / '.quilt']
```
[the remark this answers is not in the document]
ResidentMario: I didn't think to try. It does work...

api/python/t4/imports.py
```diff
@@ -0,0 +1,81 @@
...
+    Data package module loader finder. This class sits on `sys.meta_path` and returns the
+    loader it knows for a given path, if it knows a compatible loader.
+    """
+
+    @classmethod
+    def find_spec(cls, fullname, path=None, target=None):  # pylint: disable=unused-argument
+        """
+        This functions is what gets executed by the loader.
+        """
+        if 't4' not in fullname:
```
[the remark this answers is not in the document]
ResidentMario: Have to check that it starts with `t4.data` specifically, but yeah sorry for not catching this!

api/python/t4/imports.py
ResidentMario: I think you are allowed to raise exceptions inside of imports. Why would they be different from ? That being said, an `ImportError` would have been more appropriate.

That's a smarter way to handle this, though. Went ahead and made it a `finder` check.

api/python/t4/imports.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @classmethod
+    def find_spec(cls, fullname, path=None, target=None):  # pylint: disable=unused-argument
+        """
+        This functions is what gets executed by the loader.
+        """
+        # an implementation for subpackage imports exists, but this has significant
+        # consistency issues. For now let's avoid, but you can see the full code at
+        # https://github.com/ResidentMario/package-autorelaod/blob/master/loader.py
+        if not fullname.startswith('t4.data') or len(fullname.split('.')) > 3:
```
dimaryaz: Even this is a little wrong: `t4.data_transfer` will pass the check. Do `split` first, then check the first two items? E.g., `if fullname.split('.')[:2] == ['t4', 'data']`

api/python/t4/imports.py
```diff
@@ -0,0 +1,78 @@
...
+
+            # we do not know the name the user will ask for, so populate all valid names
+            for pkg in list_packages():
+                pkg_user, pkg_name = pkg.split('/')
+                if pkg_user == namespace:
+                    module.__dict__[pkg_name] = Package.browse(pkg)
+
+            module.__path__ = MODULE_PATH
+            return module
+
```
dimaryaz: Add an `else` with an `assert False`, just in case?

api/python/t4/imports.py
ResidentMario: Done, thanks. FYI, I couldn't do `startswith("t4.data.")` because we need to intercept `t4.data` as well.

PR #270, opened 2019-02-20
Support raw HTML in markdown, but sanitize it

akarve:
@nl0 you probably have a better way to do this :) -- also we probably want to forbid more tags?
@ResidentMario try this with the contents in `t4-staging/akarve/samples/`, esp `demo.md`


catalog/app/components/Markdown/Markdown.js
```diff
@@ -12,10 +13,13 @@ import { withStyles } from '@material-ui/core/styles';
 import { linkStyle } from 'utils/StyledLink';
 import * as RT from 'utils/reactTools';
 
+const SANITIZE_OPTS = {
+  'FORBID_TAGS': ['style', 'script'],
```
ResidentMario: I think we should really be using a whitelist. Theoretically with a blacklist we can miss dangerous tags in sanitization. There's also the theoretical vulnerability of new tags being added to the HTML standard that we're not aware of.

In general, I'm surprised how manual this process is. I was pretty sure that there are HTML sanitization libraries that do the right thing in one function, without requiring you to specify the tags you don't want to allow yourself.

nl0: i second this, we should use a whitelist (a short one)

nl0: @ResidentMario judging by the docs, this lib is pretty awesome and it does the necessary xss protection by default, so these forbidden tags and attrs are extra

catalog/app/components/Markdown/Markdown.js
```diff
@@ -136,7 +140,11 @@ export const getRenderer = memoize(({
     disable: !images,
     process: processImg,
   }));
-  return md;
+  const purify = createDOMPurify(window);
+  // sanitize the HTML
+  this.render =  (data) => purify.sanitize(md.render(data), SANITIZE_OPTS)
```
nl0: this is not a constructor, so `this` points to the `global` object here, which is not what you want, i guess

nl0:
@akarve i've adjusted this a little. good to go now

nl0:
@akarve 
> you probably have a better way to do this :)

this looks perfectly fine, actually

> also we probably want to forbid more tags?

i'm not sure it's necessary, we're generally safe with this setup. the most potentially dangerous thing, as i see it, is if someone wants to mess up the layout, which should be not so straightforward with disabled style attr anyway.

catalog/package-lock.json
```diff
@@ -454,7 +454,7 @@
     },
     "acorn-jsx": {
       "version": "3.0.1",
-      "resolved": "http://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
+      "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
```
nl0: @akarve we should probably sync our npm versions / configs to avoid these meaningless diffs

akarve: OK, I'm on 6.5.0. now

PR #275, opened 2019-02-22
Avoid overwriting existing configs

meffij:


akarve:
seems hacky - instead of checking for an empty bucket we should check whether one or more files in a list we maintain are in the bucket.

meffij:
When would we want to write configs into a non-empty config bucket? I would prefer to be cautious here. We could check for the presence of the files we intend to write, but I think it's cleaner and safer to just refuse to write anything into a non-empty bucket

meffij:
Closed in favor of using `navigator-config` to do schema updates on existing configs

PR #277, opened 2019-02-22
Search pagination

nl0:
Paginate search results (10 per page). The UI was taken from Q2 without modifications and may be adjusted if necessary.


nl0:
@akarve search results are now paginated. tell me if you want the controls to look differently.

nl0:
@akarve fixed now

PR #279, opened 2019-02-23
Use configured navigator_url for Bucket.search

meffij:


api/python/t4/util.py
```diff
@@ -300,11 +300,8 @@ def load_config():
         local_config = config_template
     return local_config
 
-def get_local_registry():
-    return load_config().get('default_local_registry', None)
-
-def get_remote_registry():
-    return load_config().get('default_remote_registry', None)
+def get_from_config(key):
+    return load_config().get(key)
```
stevededalus: Should this be `get(key, None)` or so we want it to throw if a field is missing (i.e. are they required)? Seems like they cant be required if you .have an `if not registry` check on 869.

stevededalus:
LGTM overall, one little question

api/python/t4/util.py
meffij: `x.get(y) === x.get(y, None)`

PR #286, opened 2019-02-26
Add CLI autoconfig

meffij:


api/python/tests/test_cli.py
```diff
@@ -0,0 +1,20 @@
...
+class CommandLineTestCase(TestCase):
+    """
+    Base TestCase class, sets up a CLI parser
+    """
+    @classmethod
+    def setUpClass(cls):
+        parser = create_parser()
+        cls.parser = parser
+
+class PingTestCase(CommandLineTestCase):
```
stevededalus: Maybe not named `PingTestCase` :-P

api/python/t4/main.py
```diff
@@ -25,6 +25,12 @@ def create_parser():
     logout_p = subparsers.add_parser("logout", description=shorthelp, help=shorthelp)
     logout_p.set_defaults(func=session.logout)
 
+    shorthelp = "Configure T4"
+    config_p = subparsers.add_parser("config", description=shorthelp, help=shorthelp)
+    config_p.add_argument("catalog_url", help="URL of catalog to config with",
```
stevededalus: Is there a public global catalog we can default to, or is that dead?

meffij: I think we decided that wasn't too helpful -- we want people to go through the catalog for login anyway, so they should have a catalog they can config against.

stevededalus: sgtm

PR #292, opened 2019-02-27
Allow `th` tag, other permissible commonmark HTML tags

akarve:


catalog/app/components/Markdown/Markdown.js
```diff
@@ -189,7 +216,7 @@ export const Container = RT.composeComponent('Markdown.Container',
   }),
   withStyles(() => ({
     root: {
-      overflow: 'auto',
```
nl0: why rules reordering here? i think it's a nice convention to put element's rules at the top with all the nested rules after that

akarve: OK

PR #295, opened 2019-02-27
Navigator config add registry url

meffij:
- Update autogenerated configs to include `registryUrl`
- Update existing configs key-by-key
- Modularize autogenerated configs for easier testing


lambdas/navigator_config/index.py
```diff
@@ -63,14 +80,55 @@ def config(dest_bucket, dest_dir, es_url, api_url, s3_bucket,
             'accessKeyId': '',
             'secretAccessKey': ''
         },
+        'registryUrl': 'https://' + registry_host + '/',
```
stevededalus: Might want to be careful about possible trailing slash in case user enters that as a parameter to the CFT.

meffij: Would it be an acceptable alternative to document in the parameter that there should be no leading or trailing slashes?

lambdas/navigator_config/index.py
```diff
@@ -14,16 +15,40 @@ def handler(event, context):
...
+def print_json(source):
+    """
+    Pretty-print config as JSON.
+    """
+    return json.dumps(source, sort_keys=True, indent=4, separators=(',', ': '))
+
+def update_config(existing, autogenerated):
+    """
+    Update existing config with new autogenerated config.
+
```
stevededalus: Add a comment in the docstring that it updates keys based on top level in json and not nested keys. I think its fairly clear since you named it update instead of merge, but might as well be explicit

lambdas/navigator_config/index.py
```diff
@@ -63,14 +80,55 @@ def config(dest_bucket, dest_dir, es_url, api_url, s3_bucket,
...
+        """
+        try:
+            response = s3_client.get_object(Bucket=dest_bucket, Key=key_name)
+            return json.load(response['Body'])
+        except ClientError as e:
+            if e.response['Error']['Code'] == "NoSuchKey":
+                return {}
+            raise e
+
+    region = boto3.session.Session().region_name
```
stevededalus: is `boto3.session.Session().region_name` the region that the current aws profile/cli is config'ed to? I'm worried that this might override a region if used against a bucket in a different region than the default configed one, but am not sure how this is used.

When does this actually get run?

meffij: that gets run by the lambda task, which will be in the same region as the stack, which is what we want.

stevededalus: cool, thanks for clarifying

lambdas/navigator_config/index.py
stevededalus: Sure, looks like we might be able to set `AllowPattern` on the registry host in the template and disallow trailing slash and http(s):// prefix ?

stevededalus: (not blocking, we can do that later)

akarve: my suggestion is that any URLs should not have a trailing slash. it's the caller's responsibility to terminate the path. that's what we do in catalog.

@nl0 do you agree? you may have seen in slack that catalog is currently calling `https://url//some/path` -- it works but not pretty

meffij: let's backlog that

akarve: re `AllowPattern` - sure, but we should first enforce that in the schema itself, via [`pattern`](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html)

nl0: @akarve yep, i agree, and catalog expects `registryUrl` with no trailing slash

meffij: I removed the trailing `/` from the autogenerated registryUrl

PR #297, opened 2019-02-28
Add a default dest to fetch

ResidentMario:
`t4.PackageEntry.fetch` and `t4.Package.fetch` will now default to downloading entries to the local equivalent of the package entry name.

In other words, after this PR, the following will be true:
* `p["foo.txt"].fetch("foo.txt") == p["foo.txt"].fetch()`
* `p["somedir/"].fetch("./") == p["somedir/"].fetch()`

This allows the user to omit the `dest` argument in *most* cases, as an exact name-to-name copy of the preexisting file is usually what you want. That's a significant UX benefit.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -439,7 +439,7 @@ class Package(object):
             pkg = pkg._children[key_fragment]
         return pkg
 
-    def fetch(self, dest):
+    def fetch(self, dest='./'):
         """
         Copy all descendants to `dest`. Descendants are written under their logical
         names _relative_ to self.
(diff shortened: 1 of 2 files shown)


ResidentMario:
This actually isn't as simple as I thought it would be.

* It requires that `PackageEntry` objects be aware of their own name, which they currently aren't. We _could_ make that an attribute of `PackageEntry`, for this specific purpose, though.
* It requires that `Package` objects be aware of their own file position on disk, which they currently aren't. Again we _could_ make that an attribute of `PackageEntry`, for this specific purpose.

ResidentMario:
I'm not sure what my previous comment was on about, because upon revisiting this idea and this PR I found that I could get this exact behavior working pretty easily.

ResidentMario:
A bad merge and manual fixes to `test_packages.py` later, and codecov is [totally lost](http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=366433).

PR #299, opened 2019-03-01
Reveal cause of 500 errors in @api decorator

akarve:


lambdas/shared/t4_lambda_shared/decorator.py
```diff
@@ -15,10 +15,10 @@ def wrapper(event, _):
             headers = event['headers'] or {}
             try:
                 status, body, response_headers = f(params, headers)
-            except Exception:
+            except Exception as ex:
                 traceback.print_exc()
                 status = 500
-                body = 'Internal Server Error'
+                body = str(ex)
```
dimaryaz: Generally, it's considered to be a bad idea to give away such info to the user... Though I can't think of anything particularly sensitive in our case, so whatever.

akarve: we can be selective about if/when we show it to the user. to take one example, a badly formatted parquet file, it'd really be nice if catalog says "corrupted file" or something like that. also helps debugging.

PR #310, opened 2019-03-06
Don't HEAD every object when installing packages

dimaryaz:
We already have sizes in the package info, so just use them.

This means, `copy_file_list` now requires the caller to pass in the size, which is a bit of a strange API - but so far, the only two callers already know the sizes.


api/python/t4/data_transfer.py
```diff
@@ -455,13 +455,12 @@ def _looks_like_dir(s):
 
 def copy_file_list(file_list):
     """
-    Takes a list of tuples (src, dest, override_meta) and copies them in parallel.
+    Takes a list of tuples (src, dest, size, override_meta) and copies them in parallel.
```
ResidentMario: When you call `t4.Package.set`, the `size` attribute of the `PackageEntry` is initialized with the value of `None`.

So imagine you create a package for the first time. E.g. `p = t4.Package().set("foo", "foo")`. If you ask for `p['foo'].size`, you will get `None`. Now let's say you push it: `p.push("a/b", "s3://whatever")`.

In this PR, this `push` will call `t4.Package._materialize` as a subroutine. `_materialize` will, in turn, call `copy_file_list` on a list of package data tuples. These tuples will set `entry.size` to the `PackageEntry.size` value. But per the above, `PackageEntry.size` is still `None`!

I don't think you can do this.

ResidentMario: This is not true of entries living on S3. In that case the package entry will point to a specific object version, and that object version is guaranteed not to change in size in between `set` time and `push` time.

I think (correct me if I'm wrong) the time cost for determining the size of local files is small enough to be negligible. So perhaps we could do this:

* Use the package-cached size values for remote entries.
* Continue to check file sizes for local entries at push time.

dimaryaz: Interesting. Technically, we already have a similar bug when calculating the hashes: `data_transfer.calculate_sha256` shows a progress bar, and it'll be incorrect if the file size has changed.

In any case, I think we'll need to set some requirements for the user - e.g., once you call `set(...)`, you don't modify the file. Otherwise, in the extreme case, the user could modify the file while it's being hashed or uploaded - and that will of course produce incorrect results.

dimaryaz: Another thing to consider: originally, we wanted to hash the file in the `set(...)` call, but decided it was too annoying, so we delayed hashing until later. It's supposed to be an optimization - so arguably, modifying the file after the `set(...)` should not be supported since it results in a different behavior. (It would be nice to catch that case... but the most we can do is check size and last modified time, I guess.)

ResidentMario: > modifying the file after the set(...) should not be supported since it results in a different behavior

In my opinion, we should warn the user when we detect that they are doing this (using the size check), update the size, and continue the push anyway. That is a separate issue that I think is material to a separate PR, which I will write out as a separate asana task.

ResidentMario: So I don't have any objections to this PR, I think.

ResidentMario:
@dimaryaz I've written up a follow-up task for the additional data model adjustment we should make once we merge this PR: https://app.asana.com/0/835394562528972/1113402836067910.

PR #312, opened 2019-03-07
Implement a thumbnail lambda

dimaryaz:
It returns a JSON response with a base64-encoded thumbnail.


lambdas/thumbnail/tests/test_thumbnail.py
```diff
@@ -0,0 +1,57 @@
...
+    def _make_event(cls, query, headers=None):
+        return {
+            'queryStringParameters': query or None,
+            'headers': headers or None
+        }
+
+    @responses.activate
+    def test_thumbnail(self):
+        url = 'https://example.com/penguin.jpg'
+        image = BASE_DIR / 'penguin.jpg'
```
akarve: traveling == QA research

lambdas/thumbnail/index.py
```diff
@@ -0,0 +1,92 @@
...
+    (128, 128),
+    (256, 256),
+    (480, 320),
+    (640, 480),
+    (960, 640),
+    (1024, 768),
+    (2048, 1536)
+]
+# Map URL parameters to actual sizes, e.g. 'w128h128' -> (128, 128)
+# TODO(dima): Is this a good idea? Don't know why Dropbox does it that way.
```
akarve: alternatives require parsing strings from the request, and that's strictly worse in performance and readability. it's kinda weird that `f'W{w}H{h}` is not supported but, meh, ppl need call the API right.

lambdas/thumbnail/index.py
```diff
@@ -0,0 +1,92 @@
...
+        'size': {
+            'enum': list(SIZE_PARAMETER_MAP)
+        }
+    },
+    'required': ['url', 'size'],
+    'additionalProperties': False
+}
+
+@api(cors_origins=ALLOWED_ORIGINS)
+@validate(SCHEMA)
```
akarve: If the client passes a bad size this will 400... that's fine but it would be nice to have a message: "Unsupported size {whatever} requested. Supported sizes are {whatever}."

lambdas/thumbnail/index.py
akarve: you could `.lower()` the string before looking it up

lambdas/thumbnail/index.py
```diff
@@ -0,0 +1,92 @@
...
+    size = SIZE_PARAMETER_MAP[params['size']]
+
+    resp = requests.get(url)
+    if resp.ok:
+        image_bytes = BytesIO(resp.content)
+        with Image.open(image_bytes) as image:
+            orig_format = image.format
+            orig_size = image.size
+            image.thumbnail(size)
+            thumbnail_size = image.size
```
akarve: since the above 4 variables are only used in the response, why even declare them? might be cleaner to do:
```
  'original_format': image.format
```
etc.

dimaryaz: The image gets modified in-place when the thumbnail is created - so need to save those variables

akarve: ah yeah

lambdas/thumbnail/index.py
dimaryaz: Actually, the JSON schema error message is pretty helpful (if a bit verbose). E.g., passing "foo" results in:
```
'foo' is not one of ['w32h32', 'w64h64', 'w128h128', 'w256h256', 'w480h320', 'w640h480', 'w960h640', 'w1024h768', 'w2048h1536']

Failed validating 'enum' in schema['properties']['size']:
    {'enum': ['w32h32',
              'w64h64',
              'w128h128',
              'w256h256',
              'w480h320',
              'w640h480',
              'w960h640',
              'w1024h768',
              'w2048h1536']}

On instance['size']:
    'foo'
```

lambdas/thumbnail/index.py
dimaryaz: Ok. Though for whatever map we pick, I think being strict (i.e., case-sensitive) is better. The other URL parameters we use - "ipynb" in the preview lambda, etc. - are case sensitive.

lambdas/thumbnail/index.py
akarve: Oh nice I should have noticed that

lambdas/thumbnail/index.py
akarve: yep agreed

dimaryaz:
@nl0: yeah, base64 is not ideal - just easiest solution right now. Binary data requires some extra API Gateway config - I'll look into it later.

We should also gzip all our JSON responses - that would make base64 less painful. Though that also requires API Gateway config.

PR #313, opened 2019-03-07
Add tokens to navigator-config

meffij:


lambdas/navigator_config/index.py
```diff
@@ -16,7 +18,7 @@ def handler(event, context):
         config(props['DestBucket'], props['DestDir'], props['ConfigEsUrl'],
```
akarve: this function is a bug waiting to happen. we should be passing in a dict instead of ten positional arguments that can be mixed up

akarve: kwargs would also be OK

meffij: will fix next week

meffij:
Closed on account of being superseded by #315

PR #314, opened 2019-03-08
Bucket setup lambda

meffij:


lambdas/bucket-setup/index.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+
+def set_cors(bucket, catalog_host):
+    new_cors_rule = {
+        'AllowedHeaders': [ '*' ],
+        'AllowedMethods': [
+            'GET',
+            'HEAD',
+            'PUT',
+            'POST'
```
kevinemoore: Intentionally leaving out DELETE?

meffij: the CORS policy Aneesh specified left out DELETE. I assume we don't support any DELETEs from the catalog at the moment, and it's better to be safe I guess

PR #315, opened 2019-03-08
Navigator config stack create

meffij:
Supersedes #313


lambdas/navigator_config/index.py
```diff
@@ -1,24 +1,43 @@
...
+        # stack create
+        try:
+            create_configs(**props)
+            cfnresponse.send(event, context, cfnresponse.SUCCESS, {})
+            return
+        except Exception:
+            cfnresponse.send(event, context, cfnresponse.FAILED, {})
+            raise
+
+    # event['RequestType'] == 'Update'
```
stevededalus: Turn this into a sentences so it's clear that you're documenting and this isn't an artifact of debugging
```
# In other cases try and update, i.e. when event['RequestType'] == 'Update'
```

lambdas/navigator_config/index.py
```diff
@@ -45,20 +64,18 @@ def update_config(existing, autogenerated):
...
     """
     bucket_config = {
-        'name': s3_bucket,
-        'title': bucket_title,
-        'icon': bucket_icon,
-        'description': bucket_description,
-        'searchEndpoint': es_url,
-        'apiGatewayEndpoint': api_url,
-        'region': region
+        'name': kwargs['ConfigS3Bucket'],
```
stevededalus: What's the logic on switching to `kwargs` here? Seems like we lose out on having the function signature without much gain. It's still going to fail if argument is missing, so doesn't make the schema more flexible. I guess it lets the template pass extra arguments from a new version to and old lambda that will ignore them and not crash, but the caller is the same CFT that references the lambda hash.  What am I missing?

meffij: Moving away from positional arguments was Aneesh's idea, and I think it's a good idea too. You can't just have 10+ positional parameters without someone screwing it up at some point.

Compatibility between mismatched templates and navigator-config's is not something I am concerned about. We don't support mismatched versions, so whether they work or not is irrelevant. The point of the change to kwargs is code quality, not a larger compatibility matrix.

stevededalus: I think you can get that effect with this and still maintain the docstrings/funtion signature:
```
autogenerate_configs(
        bucket_title=None,
        bucket_icon=None,
        ...
        region=None)
```

PR #316, opened 2019-03-09
 Handle case of no existing CORS config

akarve:


akarve:
valid concern - i will backlog

PR #317, opened 2019-03-09
Fix a unittest

dimaryaz:
Should probably add it to CI


akarve:
Hmmm... `master` passes `pytest` in `lambdas/shared` for me, but your branch fails. What am I doing wrong?

akarve:
It looks like you want to open this against #318

akarve:
Weird. Works for me now. I need this for another PR, so merging.

PR #324, opened 2019-03-12
Admin/Users: admin decoration: asterisk -> shield icon

nl0:
Use shield icon instead of asterisk for decorating admins in the user list view


nl0:
@akarve as you requested

nl0:
admin-shields

@akarve is it ok to make the icons "hanging" to keep usernames aligned (see the screenshot) or should i instead align the contents of the cell, so the icons "push" the usernames?

PR #325, opened 2019-03-12
Catalog maintenance

nl0:
## Summary

- upgrade Material UI and adjust for changes
- remove some legacy stuff
- change theme color according to new mocks


nl0:
@akarve take a look pls when you have some time

PR #326, opened 2019-03-13
WIP: Add Bucket Wiping lambda to delete objects in bucket

stevededalus:
Iterate through every object in a bucket and delete it. Used for removing the config bucket on stack delete.

~~Open question: Should config bucket be unversioned or should this iterate over versioned objects?~~
Open question: Best way to write tests against lambdas?

### DIFF
--- a/lambdas/bucket-wiper/index.py
+++ b/lambdas/bucket-wiper/index.py
@@ -13,9 +13,12 @@ def lambda_handler(event, context):
         if event['RequestType'] == 'Delete':
             s3 = boto3.resource('s3')
             bucket = s3.Bucket(bucket)
-            for obj in bucket.objects.filter():
-                s3.Object(bucket.name, obj.key).delete()
-
+            # for obj_version in bucket.object_versions.all():
+            versioned_objs = []
+            for obj_version in bucket.object_versions.all():
+                versioned_objs.append({'Key': obj_version.object_key,
+                                       'VersionId': obj_version.id})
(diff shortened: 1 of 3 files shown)


lambdas/bucket-wiper/index.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
```
dimaryaz: Don't need this - that's not an executable script

lambdas/bucket-wiper/index.py
```diff
@@ -0,0 +1,25 @@
...
+
+        if event['RequestType'] == 'Delete':
+            s3 = boto3.resource('s3')
+            bucket = s3.Bucket(bucket)
+            # for obj_version in bucket.object_versions.all():
+            versioned_objs = []
+            for obj_version in bucket.object_versions.all():
+                versioned_objs.append({'Key': obj_version.object_key,
+                                       'VersionId': obj_version.id})
+            bucket.delete_objects(Delete={'Objects': versioned_objs})
```
dimaryaz: It supports up to 1000 objects at a time - so need to do it in chunks

stevededalus: Hopefully we won't have 1000 objects in the config bucket, but updated just in case.

stevededalus:
Successfully created and destroyed:
quilt-t4-staging-d5a4c0

PR #327, opened 2019-03-15
Add meta support to put_file

meffij:


api/python/t4/bucket.py
```diff
@@ -134,13 +134,15 @@ def put(self, key, obj, meta=None):
...
-    def put_file(self, key, path):
+    def put_file(self, key, path, meta=None):
         """
         Stores file at path to key in bucket.
 
         Args:
             key(str): key in bucket to store file at
             path(str): string representing local path to file
+        Optional args:
+            meta(dict): metadata to attach to file
```
stevededalus: Do we end up putting this in the `user metadata` section of the S3 metadata? It looks like this is still using the `helium` metadata key
Should document the limit on metadata size here too (2K?)

PR #329, opened 2019-03-16
Add generic TXT preview, improve unit tests, don't return HTML for VCF or TXT

akarve:


lambdas/preview/index.py
```diff
@@ -163,13 +168,61 @@ def extract_vcf(file_):
...
+                ellipsis = True
+                break
+    # if maxlines == 0, everything still works
+    available = min(available, MAX_LINES)
+    # pylint should never flag for 3.6 but it does
+    headlines = int(available/2) # pylint: disable=old-division
+    taillines = available - headlines
+
+    head = subprocess.check_output(['head', file_.name, '-n', str(headlines)])
+    tail = subprocess.check_output(['tail', file_.name, '-n', str(taillines)])
```
stevededalus: You can use SEEK_END to scan from the end of the file.
```
    f.seek(-2, os.SEEK_END)     # Jump to the second last byte.
    while f.read(1) != b"\n":   # Until EOL is found...
        f.seek(-2, os.SEEK_CUR) # ...jump back the read byte plus one more.
    last = f.readline()    
```
And for head, I think you don't need anything special, you can just open the file and read the first n lines as normal.

lambdas/preview/index.py
```diff
@@ -163,13 +168,61 @@ def extract_vcf(file_):
...
+    html_buffer = [[], []]
+    for i, lines in enumerate([head, tail]):
+        for line in lines.splitlines():
+            line = line.strip().decode('unicode_escape')
+            html_buffer[i].append(f'<p>{line}</p>')
+
+    # pylint: disable=bad-continuation
+    html = ('<div>'
```
stevededalus: Should, pick a consistent quote to use. Looks like maybe double quotes for python strings and single quotes withint the strings for html values?

lambdas/preview/index.py
```diff
@@ -163,13 +168,61 @@ def extract_vcf(file_):
...
+                "<div data-type='head'>{}</div>"
```
stevededalus: do you want the indents inside the strings?

lambdas/preview/index.py
akarve: single quote is PEP-8 default but when i have nested quotes then double quote becomes necessary

lambdas/preview/index.py
akarve: so developer can see HTML indentation but this whole HTML thing is busted and i'm going to refactor

lambdas/preview/index.py
akarve: wanted to avoid but looks easy enough; let me see if i can use this

lambdas/preview/index.py
stevededalus: Yeah, I meant do you want the indents in the html too so that it looks ok in the dev tools, like:
```
"    <div.."
```
vs
```
    "<div.."
```

If you don't care about the readability/indentation of the out put htlm you could wrap the whole thing in triple quotes instead.

lambdas/preview/index.py
stevededalus: Actually PEP-8 and Google Style don't default to singles. I'd just swap them all to double so that the single quotes inside don't have to be escaped

> String Quotes
> In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it. When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.
https://www.python.org/dev/peps/pep-0008/#string-quotes

akarve: we're returning JSON now so doesn't matter, but noted for future

lambdas/preview/index.py
akarve: i shouldn't have returned HTML in the first place :P (unnecessarily emulating what we did with nbformat). we return JSON now so this is a non-issue, but the problem with newlines in the HTML text is that they get encoded as `\n` and show up in the DOM. an AST or fully server-side-rendered is preferred.

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
-    return html, {}
+def _truncate(line, size):
+    """chop string, if needed, to fit in remaining bytes
+    Args:
+        line - string to truncate
+        size - bytes consumed so far (without input_string)
+    """
+    remaining = MAX_BYTES - size
+    if 0 < remaining < len(line):
+        line = line[:remaining]
```
stevededalus: A char isn't necessarily a byte at this point

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
 
-    return html, {}
+def _truncate(line, size):
+    """chop string, if needed, to fit in remaining bytes
+    Args:
+        line - string to truncate
+        size - bytes consumed so far (without input_string)
+    """
+    remaining = MAX_BYTES - size
+    if 0 < remaining < len(line):
```
stevededalus: length here is length in unicode chars and not bytes.

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
     header = []
     data = []
-
+    size = 0
     with open(file_.name, 'rb') as vcf:
         for index, line in enumerate(vcf, start=1):
-            # don't escape quotes
-            line = line.strip().decode('unicode_escape')
-            formatted = f'<p>{line}</p>'
+            line = _truncate(line.rstrip().decode(), size)
```
stevededalus: add an explicit decoding scheme as an argument to decode. 

I'm also a little surprised that rstrip works on a bytes object and that it shouldn't be `line.decode().rstrip()` instead

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
-            "</div>").format(
-                ''.join(meta) + '<br />',
-                ''.join(header) + '<br />',
-                ''.join(data))
 
-    return html, {}
+def _truncate(line, size):
+    """chop string, if needed, to fit in remaining bytes
+    Args:
+        line - string to truncate
```
stevededalus: if you care about bytes instead of chars, might want to just leave line as `bytes`

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
+                meta.append(line)
             elif line.startswith('#'):
-                header.append(formatted)
+                header.append(line)
             else:
-                data.append(formatted)
+                data.append(line)
+            size += len(line)
+            # stop if we're over the max
+            if index >= MAX_LINES or size > MAX_BYTES:
```
stevededalus: same issue with chars vs bytes

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
+
+def extract_txt(file_):
+    """
+    Display first and last few lines of a potentially large file.
+
+    Args:
+        file_ - file-like object
+    """
+    size = 0
+    ellipsis = False
```
stevededalus: Throw an end of line comment here like
```
ellipsis = False # File long enough to need ellipsis
```
or use a more boolean name like `needs_ellipsis`

lambdas/preview/index.py
stevededalus: also, this line is a little confusing because it switched the type of line via a function

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
+    head = []
+    with open(file_.name, 'rb') as txt:
+        for index, line in enumerate(txt, start=1):
+            line = _truncate(line.rstrip().decode(), size)
+            size += len(line)
```
stevededalus: same `bytes` vs `string` weirdness here.

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
+            # this is a heuristic; can fail to return shorter tail lines
+            # if head lines are uncharacteristically long
+            # we're guarding against, for example, huge single-line JSON files
+            if index <= MAX_LINES and size <= MAX_BYTES:
+                head.append(line)
+            if size > MAX_BYTES:
                 break
+            if index > MAX_LINES:
+                ellipsis = True
```
stevededalus: since this only happens in one case, you don't need a flag variable for ellipsis, you can just inline that whole method here and you don't need to reopen the file, immediately seek to the end and start your while loop. If you're worried about it being too much indentation you can use a named inner function instead (but cutting the `if ellipsis` and `with open` out means it shouldn't be too bad).

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
     """
     Pull summary info from VCF: meta-information, header line, and data lines
     (in that order, up to MAX_LINES)
-    """
 
+    Args:
+        file_ - file-like
+    """
```
stevededalus: Throw a returns string in here. Not usually a big deal, but with all the string vs bytes going on it helps.

lambdas/preview/index.py
dimaryaz: Can we just open the file as `'r'` instead of `'rb'`? Or is byte counting that important?

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
+            if index <= MAX_LINES and size <= MAX_BYTES:
+                head.append(line)
+            if size > MAX_BYTES:
                 break
+            if index > MAX_LINES:
+                ellipsis = True
+                break
+    tail = []
+    if ellipsis: # in this case we need a tail
+        headlen = int(MAX_LINES/2) # pylint: disable=old-division
```
dimaryaz: Why not just `MAX_LINES // 2`?

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
+        taillen = MAX_LINES - headlen
+        # cut head
+        head = head[:headlen]
+        # grow tail
+        with open(file_.name, 'rb') as txt:
+            count = 0
+            txt.seek(0, os.SEEK_END) # the very end
+            while count < taillen:
+                txt.seek(-2, os.SEEK_CUR)
```
dimaryaz: This seems crazy inefficient - seek and read, one byte at a time?

lambdas/preview/index.py
dimaryaz: But, truncating bytes could break UTF8 characters. If we really want to be that precise, we'd need to decode the truncated string with `decode('utf8', 'ignore')`

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
-                "<div data-type='meta'>{}</div>"
-                "<div data-type='header'>{}</div>"
-                "<div data-type='data'>{}</div>"
-            "</div>").format(
-                ''.join(meta) + '<br />',
-                ''.join(header) + '<br />',
-                ''.join(data))
 
-    return html, {}
+def _truncate(line, size):
```
dimaryaz: This function looks weird to me. I'd rather pass in the remaining bytes. Then, the whole function could just be:
```
return line[:max(remaining, 0)]
```

lambdas/preview/index.py
akarve: is there a faster way to find the lines at the end of a file when we don't know line width?

akarve: oh - we could read it into memory in large chunks, sure

dimaryaz: Ideally, read from the end in chunks of a few KBs.

Don't go too crazy optimizing this, though: even reading every single line of the file wouldn't be that big of a deal, given that we just downloaded the whole file from S3. But making two syscalls per byte could end up pretty expensive if you have a single really long line.

Technically, though, even reading a single line isn't safe: theoretically, you could have a GB file with no line breaks.

lambdas/preview/index.py
```diff
@@ -141,35 +145,102 @@ def extract_vcf(file_):
...
+    """
+    Display first and last few lines of a potentially large file.
+
+    Args:
+        file_ - file-like object
+    """
+    size = 0
+    ellipsis = False
+    head = []
+    with open(file_.name, 'rb') as txt:
```
dimaryaz: Do you actually need to open it again? `file_` is already an open file.

Same for the VCF function.

lambdas/preview/index.py
akarve: we know how many bytes are left in the budget so i can just seek back that far and get as many whole lines as are available

dimaryaz: Oh yeah, that works

lambdas/preview/index.py
akarve: byte strings better for size counting. doc-string notes that, in some rare cases (e.g. a file on a really long line) we may break a unicode character mid-stream. but usually this is not an issue (we don't usually break lines up).

akarve: actually... i think we can

akarve: i looked into this and it 1) breaks the parquet unit test (where we don't even touch encoding); 2) breaks all but one other test and, at first glance, did not find a simple fix.

lambdas/preview/index.py
```diff
@@ -34,15 +36,17 @@
...
+    Returns:
+        JSON response
     """
     url = params['url']
     input_type = params.get('input')
 
     resp = requests.get(url)
     if resp.ok:
-        with NamedTemporaryFile() as file_:
+        with NamedTemporaryFile(mode='r+b') as file_:
```
dimaryaz: Does `'r+b'` make a difference?

lambdas/preview/index.py
```diff
@@ -141,35 +147,95 @@ def extract_vcf(file_):
...
+        if line and len(head) < MAX_LINES:
+            head.append(line)
+        # if the file is longer than MAX_LINES, we need a tail
+        if index > MAX_LINES:
+            headmax = MAX_LINES//2
+            tailmax = MAX_LINES - headmax
+            # cut the head back to make room for the tail
+            head = head[:headmax]
+            # correct the size for the lines we just dropped
+            size = sum([len(s) for s in head])
```
dimaryaz: Don't need `[` and `]` - that just builds a list unnecessarily

lambdas/preview/index.py
```diff
@@ -141,35 +147,95 @@ def extract_vcf(file_):
...
+            file_size = os.stat(file_.name).st_size
```
dimaryaz: `os.fstat(file_.fileno())` or `file_.seek(0, os.SEEK_END)` would be better. In general, a temporary file might not even have a name. (It does in this particular case cause the Parquet parser needs it, and we're using `NamedTemporaryFile`, but it's good to not rely on that.)

lambdas/preview/index.py
```diff
@@ -141,35 +147,95 @@ def extract_vcf(file_):
...
+            # avoid rewinding past start of file
+            remaining = min(MAX_BYTES - size, file_size)
+            # go to the earliest available byte
+            file_.seek(-remaining, os.SEEK_END)
+            tail = [l for l in file_.read().split(b'\n') if l]
```
dimaryaz: Why not `[l for l in file_ if l]`?

lambdas/preview/index.py
akarve: yeah fixed - good feedback. all of the loops got tighter and shorter, so the local function declaration was not even needed

lambdas/preview/index.py
akarve: oh you're right. it _should_ but it doesn't even work.

lambdas/preview/index.py
```diff
@@ -140,36 +146,97 @@ def extract_parquet(file_):
 def extract_vcf(file_):
     """
     Pull summary info from VCF: meta-information, header line, and data lines
-    (in that order, up to MAX_LINES)
-    """
+    (in that order, up to MAX_LINES). Skips empty lines.
 
+    Args:
+        file_ - file-like object
```
stevededalus:
```
file_ - file-like object opened in binary mode (+b)
```

lambdas/preview/index.py
```diff
@@ -140,36 +146,97 @@ def extract_parquet(file_):
...
-    html = ("<div>"
-                "<div data-type='meta'>{}</div>"
-                "<div data-type='header'>{}</div>"
-                "<div data-type='data'>{}</div>"
-            "</div>").format(
-                ''.join(meta) + '<br />',
-                ''.join(header) + '<br />',
-                ''.join(data))
+    Args:
+        file_ - file-like object
```
stevededalus:
```
file_ - file-like object opened in binary mode (+b)
```

PR #334, opened 2019-03-19
Bucket/Dir: refactor and cleanup

nl0:
trivial refactoring. rtm.


nl0:
@akarve this one is trivial, so ima merge it

PR #335, opened 2019-03-19
Catalog: restore Intercom

nl0:
to enable intercom just add `intercomAppId` key to the config


akarve:
Thanks; I've also added the key to config-schema.json

PR #338, opened 2019-03-20
Fix cli config

meffij:


api/python/t4/main.py
```diff
@@ -29,7 +28,7 @@ def create_parser():
     config_p = subparsers.add_parser("config", description=shorthelp, help=shorthelp)
     config_p.add_argument("catalog_url", help="URL of catalog to config with",
```
dimaryaz: Can't you just rename it here - call it "autoconfig_url"? Or even leave the name, but add `dest='autoconfig_url'` to rename the argument to `config`.

meffij: it looks like the way Brian wrote `config`, any named parameters will just be added to the config, including `autoconfig_url`. The only way to autoconfig with `config` is to use exactly one positional parameter.

PR #341, opened 2019-03-22
Catalog: add option to disable signup

nl0:
via `disableSignUp` config key (boolean, optional)


akarve:
For future PRs, you should also submit an update to config-schema.json in deployment so that we stay on top of these changes

PR #346, opened 2019-03-23
Catalog: update preview logic (VCF, text, Excel)

nl0:
## Summary

- [x] preview VCF files
- [x] use preview endpoint for text formats
- [x] preview CSV/TSV
- [x] use thumbnailer for images
- [x] support compression (`.gz`, `.bz2`)
- [x] get and parse `Content-Range` header

### DIFF
--- a/catalog/app/components/Preview/Preview.js
+++ b/catalog/app/components/Preview/Preview.js
@@ -19,6 +19,7 @@ const loaderChain = [
   loaders.Markdown,
   loaders.Notebook,
   loaders.Parquet,
+  loaders.Vcf,
   loaders.Text,
   loaders.Image,
   fallback,
--- a/catalog/app/components/Preview/Preview.js
+++ b/catalog/app/components/Preview/Preview.js
@@ -15,6 +15,7 @@ const fallback = {
 };
 
 const loaderChain = [
+  loaders.Csv,
   loaders.Json,
   loaders.Markdown,
   loaders.Notebook,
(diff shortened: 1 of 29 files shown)


catalog/app/components/Preview/Preview.js
```diff
@@ -15,10 +15,12 @@ const fallback = {
 };
 
 const loaderChain = [
+  loaders.Csv,
```
akarve: fine for now, but i question the casing. we have csv, CSV, Csv, etc. all in the same combined codebase

catalog/app/components/Preview/Preview.js
```diff
@@ -15,10 +15,12 @@ const fallback = {
...
   loaders.Json,
   loaders.Markdown,
   loaders.Notebook,
   loaders.Parquet,
+  loaders.Vcf,
```
akarve: OK for separate PR, but we now support XLS and XLSX as well - works like .parquet (minus the info metadata)

catalog/app/components/Preview/Preview.js
nl0: we use the js CamelCase (not sure if it's 100% precise term) convention here, abbreviations are considered words, so only the first letter should be capitalized, e.g. `JsonEncoder`. it's not 100% consistent across the codebase, but i'm trying to adhere to that convention.

catalog/app/components/Preview/Preview.js
nl0: ok

PR #359, opened 2019-03-30
Update and cleanup re: search

eode:
## Description
The largest effect of this are:
 * `api.search()` has been removed
 * `api.log()` has been removed
 * `Bucket.search()` now works.
 * `api.config()` once again retains the catalog used in configuration.

Key changes are in `search_util.py` and `api.py`.

Other (minor or cosmetic) changes:
 * HeliumConfig renamed to T4Config
 * T4Config no longer parses the config template unless it needs to
 * `autoconfig_url` renamed to `catalog_url` to make it more clear to the end user when calling `api.config()`

Note:  The test added may be superfluous -- since it's off of a Bucket object, I added it to `test_bucket.py`, however, someone had added bucket-specific test test in `test_search.py`, and the tests are fairly similar -- though, mine did draw out a region-related issue.

It should be noted that Calvin did the groundwork for T4 to handle federation, I just fixed a couple bugs that were preventing it from working.


eode:
Question:  Is it the proper behavior to pass the bucket's region (as configured in federation.json) to ES for signing, or should the region passed to ES for signing be based on the ES service's region?  ..is there a potential difference?

eode:
Also, re: coverage:  I think that codecov is just unhappy with the relative change to `api.py` -- since the removed methods were covered, and some others in that file aren't, coverage for `api.py` went down, even though actual coverage didn't decrease.

eode:
Also, it should be noted that Calvin had already done the groundwork for handling federation for buckets, there were just a couple bugs that were preventing it from working.

meffij:
Regarding ES region, I think it's always correct to use what's in `federation.json`

ResidentMario:
I don't understand how to get search working, and I couldn't immediately reach an understanding of how to get it working from any of our documentation or from this code. I had to ask Calvin for clarification. 😢

Here's what I expect to work:

```python
b = t4.Bucket("s3://quilt-example")
b.search("New York")
```

I get:

```
QuiltException: Must t4.config(navigator_url=...) or specify config_url
```

At some point all of this config stuff is going to need a facelift to smooth out this experience but that's for another day...

Things I tried which failed:

```
t4.config(navigator_url="https://t4-stage.quiltdata.com/config.json")
t4.config(navigator_url="https://t4-stage.quiltdata.com/")
```

What finally worked:

```
t4.config(navigator_url="https://t4-stage.quiltdata.com")
```

Argh. At a minimum we should accept `t4.config(navigator_url="https://t4-stage.quiltdata.com/")` as well.

akarve:
@ResidentMario we can make this work any way we like, but i find this, which already works, simplest; positional args: `t4.confg(URL)`

api/python/tests/test_bucket.py
```diff
@@ -246,3 +247,25 @@ def test_bucket_config(self, config_mock, bucket_config_mock):
...
+                    "name": "t4-testing-fake",
+                    "searchEndpoint": "https://es-fake.endpoint",
+                    "region": "us-meow"
+                },
+            ]
+        }
+        self.requests_mock.add(responses.GET, 'https://foo.bar/config.json', json=content, status=200)
+        self.requests_mock.add(responses.GET, 'https://foo.bar/federation.json', json=federations, status=200)
+        b = Bucket('s3://t4-testing-fake')
+        b.search('blah', limit=1)
```
stevededalus: can you add a metadata search test?

stevededalus:
This is blowing up when using a trailing slash in the `navigator_config` parameter, but doesn't fail until searching later.

stevededalus:
Can we get this landed soon? Lots of work debugging search issues and would be helpful to have the client fixes on master

eode:
> Can we get this landed soon? Lots of work debugging search issues and would be helpful to have the client fixes on master

Should be able to land this today, and then I'll be back to Excel plugin.

eode:
> I don't understand how to get search working, and I couldn't immediately reach an understanding of how to get it working from any of our documentation or from this code. I had to ask Calvin for clarification. cry
> 
> Here's what I expect to work:
> 
> ```python
> b = t4.Bucket("s3://quilt-example")
> b.search("New York")
> ```

That works when your `t4` instance is configured to use the catalog for that bucket.

> I get:
> 
> ```
> QuiltException: Must t4.config(navigator_url=...) or specify config_url
> ```

I've updated that exception to:
"Must set t4.config(catalog_url), or set `navigator_url` directly (see t4.config docstring)"

The short and simple of it is:
t4.config() -> get current config
t4.config(catalog_url) -> set current config based on catalog's config
t4.config(navigator_url=catalog_url) -> set only the 'navigator_url' config key.

I'm happy to make changes to that if needed, and there are a few minor things that need updating in the config (such as no longer using the term 'navigator', since it's not used anymore).

> At some point all of this config stuff is going to need a facelift to smooth out this experience but that's for another day...

Yup.
> ```
> t4.config(navigator_url="https://t4-stage.quiltdata.com")
> ```
> 
> Argh. At a minimum we should accept `t4.config(navigator_url="https://t4-stage.quiltdata.com/")` as well.

Yup.  Slash will be stripped.  Should I also strip `config.json` in case someone specifies that specifically?

eode:
Wait a second -- re: navigator_url:

Does this mean that we're going to vet the values of config keys in general as they are set, or is this more a matter that the search() function should be capable of handling a trailing slash?

ResidentMario:
I want to summarize the issue with `navigator_url` in this comment for posterity.

The need to specify a `navigator_url` ahead of time, before you can actually perform searches, is from a user perspective a wart. I asked @akarve why we couldn't get rid of it, and the short answer is that that would require one of two things, neither of which works.

Recall that we have a connection to a specific S3 bucket (`s3://...`), and we do not have, and need, the URL of the specific search endpoint corresponding with that bucket (if one exists). In other words, a service discovery problem.

Here are two possible solutions:

* Writing a file in the S3 bucket to a well-known path that would tell the client what the search endpoint URL is. This is bad for a few reasons. One is that we can get name collisions, which would stop you from being able to deploy the catalog on that bucket in the first place. Another, more important issue is that this would require creating a publicly accessible object within what might otherwise be a highly private S3 bucket, which creates an uncomfortable security boundary.
* Writing a file in a separate bucket, whose name is a previously known mutation of the bucket that the user is trying to search on. However we've found from experience that creating a separate config bucket which is a non-UUID-based mutation of your actual bucket is fragile. It creates a hidden dependency on your target bucket, which is that there is no bucket named e.g. `my-bucket-quilt-config` that you don't already control (recall that S3 buckets are uniquely named). If that bucket *does* already exist, there's no consistent workaround.

Hence the need to specify a `navigator_url` before performing search in the client.

eode:
Per chat on Slack, modifying value as written to config, as opposed to having the function handle it (which seems right to me, anyways).

ResidentMario:
First some high-level comments.

A freshly created `t4.config` has the following fields. `default_local_registry`, `default_remote_registry`, and `default_install_location` are all user settings that configure how the client behaves. `navigator_url` is a catalog-facing setting that controls where the client goes to find the elastic search endpoint (and the focus of this PR).

There is also `elastic_search_url` which is vestigial; it used to be used for configuring `t4.api.search`, but is no longer necessary because this function has now been removed, so let's remove this parameter.

Finally, there is `registryUrl`, which has a hard-coded default value of `https://quilt-t4-staging-registry.quiltdata.com`. This field is used for finding the authorization registry. Let's change the naming from `registryUrl` to `registry_url`. 🙂 Yes, I know it's called `registryUrl` in `config.json`, but for naming consistency on the client let's just silently transform `registryUrl` in the catalog `config.json` into `registry_url` in the client `t4.config`.

If you run `t4.config(navigator_url="...")` you update just the `navigator_url` value, and everything works. If you run `t4.config(...)` you dump everything in the catalog's `config.json` file into your local config. That's just silly. There are a *lot* of values in the remote `config.json` file that the client doesn't care about. Instead of copying over everything, we should just copy over just the fields that matter: `navigator_url` and `registry_url`.

Reading the current code, I got the impression that we follow a different code path if we detect a `federations` field. Why do we do this?

----

One point of feedback:

> I've updated that exception to:

Let's use the following error message text, which I think is clearer:
* No search endpoint set. To set one run "t4.config(navigator_url)", which "navigator_url" is the URL of your catalog homepage.

We should encourage *one* way of doing things. And the "right" way to set a config parameter in this case is to do the full copy because that nets your `registry_url` as well, which you will probably also need and which is _much_ less discoverable on its own.

ResidentMario:
I've filed a backlog issue which is separate to this one asking for warnings on vestigial fields: https://app.asana.com/0/943271320705704/1117475802362390

api/python/t4/api.py
```diff
@@ -395,137 +395,7 @@ def sorter(pkg_info):
...
-            for t in table:
-                ptable.append((t['time'], t['version_id'], t['operation'],
-                               t['size'], t['meta']))
-            _print_table(ptable)
-        return table
-    except KeyError as e:
-        return r
-
-def config(*autoconfig_url, **config_values):
+def config(*catalog_url, **config_values):
```
stevededalus: why is `catalog_url` a splat here and not just an optional keyword?

eode:
@ResidentMario 

This has been bugging me for a while, and I just haven't had the time to go through and see what was used and what wasn't.  At the initial time of creation, the task was to automatically configure by downloading `config.json` and using it locally.  

Literally everything you're describing is drift between the Catalog and the T4 Python client, or stuff that has been missed and re-implemented in the t4 client.

* setting individual values wasn't even a part of the initial task.  I just added it because I thought it would be useful to have a way to set them locally, in case someone wanted to override the remote values.
* `config.json` changed to camelCase, but `t4`'s config is still snake_case.
* In the mean time, someone updated the client to use `registryUrl`.
* some values in `config.json` haven't ever been used on the Python side, over time, more have been added for the Catalog (navigator)'s use, but weren't useful to the Python `t4` client
* some value names have changed or moved
* it appears that for some purposes, other functions have been created and used in the mean time

WIthout going into a *ton* of further detail:  I'm in total agreement with you on this, and I really think what's needed is a bit of a refactor where configuration and state retention is concerned -- including:

* using a consistent means of accessing the config file
* storing all ephemeral state in the config if the state gets saved
* making it easy to save (or to not save) the ephemeral state
* updating the template
* applying and using the template again for something other than just default values (tracking erroneous values, making sure that the added names are known).

In any case, I think that's largely a matter of code drift and usage fragmentation, and needs its own PR (or two, or three) -- yes?

api/python/t4/api.py
eode: It was perhaps namespace paranoia on my part.

All args go into **catalog_url, and all kwargs go into **config_values.  Both are optional.  Otherwise, if the user called `config(catalog_url='blah')`, that would trigger the config update from the catalog.  Now it sets the value `catalog_url` in the config.

I would have preferred to have the interface be an object, but that was considered too heavyweight at the time.  The goal at the time was also to set the config via the catalog, not to set individual keys/values, but I thought k/v setting for the config was important.

```python
# As is
t4.config(<url)  # autoconfigure from catalog
t4.config(catalog_url=<url>)  # set the key `catalog_url` to the value of <url>

# otherwise
t4.config(<url)  # autoconfigure from catalog
t4.config(catalog_url=<url>)  # autoconfigure from catalog
```

eode:
> Let's use the following error message text, which I think is clearer:

>  No search endpoint set. To set one run "t4.config(navigator_url)", which "navigator_url" is the URL of your catalog homepage.

> We should encourage one way of doing things. And the "right" way to set a config parameter in this case is to do the full copy because that nets your registry_url as well, which you will probably also need and which is much less discoverable on its own.

Sounds good to me -- however, can I use `catalog_url` instead of `navigator_url`?  `navigator_url` is dated.

api/python/tests/test_bucket.py
eode: I missed this -- but, can you be more specific about what you want here?

There are two tests, one `test_bucket.test_search_bucket` that I wrote, and another in `test_search.test_search_bucket` that it appears Calvin wrote.  I think expanding my test would duplicate his work, and mine only covers a small amount his didn't.

If you mean specific metadata search vs plaintext, broad search, that's just a difference in the query string -- which in either case here, isn't processed anyways, because it's mocked by the time it leaves our code.

Let me know if I've missed the ball on what you want here.

PR #360, opened 2019-04-01
Correct list_package semantics when there are multiple package versions

ResidentMario:
This PR fixes and improves `t4.list_packages` so that it better handles the case of multiple package versions. Specifically, it:

* Corrects some variable naming collisions that cause all package versions to be displayed with the tag "`latest`".
* Corrects the `__iter__` logic to serve a list of packages, *not* package versions.

### DIFF
--- a/api/python/t4/api.py
+++ b/api/python/t4/api.py
@@ -214,7 +214,9 @@ def list_packages(registry=None):
         """Display wrapper for list_packages"""
 
         def __init__(self, pkg_info):
-            self.pkg_names = [info['pkg_name'].replace(':latest', '') for info in pkg_info]
+            self.pkg_names = list(set(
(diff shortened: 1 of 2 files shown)


api/python/t4/api.py
```diff
@@ -214,7 +214,9 @@ class PackageList:
         """Display wrapper for list_packages"""
 
         def __init__(self, pkg_info):
-            self.pkg_names = [info['pkg_name'].replace(':latest', '') for info in pkg_info]
+            self.pkg_names = list(set(
```
meffij: Please add a comment explaining that `list(set(foo))` is used to remove duplicate values from a list

ResidentMario: Done.

ResidentMario:
Had to fix a failed merge attempt by hand. Codecov seems broken here.

PR #368, opened 2019-04-03
Determine and set object size at set_dir time

ResidentMario:
The current `set_dir` doesn't work as expected when given a remote path because it doesn't remember to check and set object size, which is inconsistent with the behavior of `set`, which does. This is behavior that `push` now relies on, so we need to do this here, too.

In the future we should move the size check to a dedicated step in `push` and/or `build`.

### DIFF
--- a/api/python/tests/integration/test_packages.py
+++ b/api/python/tests/integration/test_packages.py
@@ -435,9 +435,9 @@ class PackageTest(QuiltTestCase):
             pkg = Package()
 
             list_object_versions_mock.return_value = ([
-                dict(Key='foo/a.txt', VersionId='xyz', IsLatest=True),
-                dict(Key='foo/x/y.txt', VersionId='null', IsLatest=True),
-                dict(Key='foo/z.txt', VersionId='123', IsLatest=False),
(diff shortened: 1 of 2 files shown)


api/python/tests/integration/test_packages.py
```diff
@@ -435,16 +435,17 @@ def test_s3_set_dir(self):
...
+                dict(Key='foo/x/y.txt', VersionId='null', IsLatest=True, Size=10),
+                dict(Key='foo/z.txt', VersionId='123', IsLatest=False, Size=10),
             ], [])
 
             pkg.set_dir('', 's3://bucket/foo/', meta='test_meta')
 
             assert pkg['a.txt'].physical_keys[0] == 's3://bucket/foo/a.txt?versionId=xyz'
             assert pkg['x']['y.txt'].physical_keys[0] == 's3://bucket/foo/x/y.txt'
             assert pkg.get_meta() == "test_meta"
+            assert pkg['x']['y.txt'].size is not None  # GH368
```
dimaryaz: Just make it `== 10`?

PR #380, opened 2019-04-05
Catalog: fix error reporting for Auth pages

nl0:


nl0:
trivial fix, merging in

PR #381, opened 2019-04-05
Catalog: CSV preview: skip lines w/ errors

nl0:
I've stumbled upon a strange preview endpoint behaviour: [see the link](http://localhost:3000/b/quilt-aics/tree/aics/label-free-imaging-collection/data_feats.csv).
The last line of the head is clipped, so the csv parser was throwing an exception, but there was no indication that the contents are clipped, and the tail was empty. I've configured parser to skip the lines with errors -- it works, but there's no indication that the line(s) are skipped, so it may mislead the user. I'd suggest handling this stuff on the backend: either returning the whole line or skipping some and putting the rest into the tail, so there's clear indication that the lines were skipped.

cc @akarve 

PS: most of the changes in this PR are styling tweaks for table display, not _directly_ related

### DIFF
--- a/catalog/app/components/Preview/loaders/Csv.js
+++ b/catalog/app/components/Preview/loaders/Csv.js
@@ -13,6 +13,7 @@ export const detect = R.pipe(utils.stripCompression,
 export const load = utils.previewFetcher('txt', ({ info: { data } }, { handle }) => {
   const opts = {
     delimiter: utils.stripCompression(handle.key).endsWith('tsv') ? '\t' : ',',
+    skip_lines_with_error: true,
   };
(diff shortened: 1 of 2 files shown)


akarve:
hmm, we shouldn't be parsing CSVs in the browser, we should rely on the endpoint to return whole lines. maybe endpoint has bug or file is corrupted

akarve:
instead of this PR I'd like to fix the preview endpoint. it has been changed and ONLY head lines are ever returned. in an upcoming PR you will have a `ilne_count=` parameter and you can request a specific number of lines.

nl0:
see #403

PR #387, opened 2019-04-08
Support tqdm progbars in environments lacking ipywidgets

ResidentMario:


api/python/t4/data_transfer.py
```diff
@@ -20,8 +20,17 @@
 
 import warnings
 with warnings.catch_warnings():
+    # suppress the TqdmExperimentalWarning returned by tqdm.autonotebook
     warnings.simplefilter("ignore")
-    from tqdm.autonotebook import tqdm
+    try:
+        from tqdm.autonotebook import tqdm
+    except ImportError:
+        warnings.warn(
```
dimaryaz: Aren't warnings being filtered out here?

ResidentMario:
I neglected to include a link to the thread where Jackson reported this issue on Slack, which is where I also wrote a little summary of the issue. [Here's that thread](https://quiltusers.slack.com/archives/C1QMGJTT2/p1554664803027500), and here's my write-up:

> `tqdm.autonotebook`, which we use to provide widget-based progress bars in the Jupyter context, and which falls back to console progress bars, has an optional dependency on `ipywidgets`. It’s possible, in theory, to have an environment which has Jupyter (or IPython?), but not the `ipywidgets` module.
> 
> Note that I can’t reproduce this issue locally. Both `conda install jupyter` and `pip install jupyter` in fresh environments work as expected for me. It’s possible it’s relying on state somewhere on my machine, though. `ipywidgets` is kind of buggy, in my experience.
> 
> Including `ipywidgets` as a dependency wouldn’t help because, since it’s a Jupyter plugin, in addition to installing it, you also need to configure it. :slightly_smiling_face:
> 
> I’ve made a PR that adds a warning in this case and falls back to using the `tqdm.tqdm` console logger if `ipywidgets` isn’t installed.

api/python/t4/data_transfer.py
ResidentMario: Yes. 🤦‍♂️

ResidentMario:
> How is it even possible to get an ImportError? I tried from tqdm.autonotebook import tqdm in a clean venv, and it just imported the regular tqdm.

I think you have to put your environment in a moderately unusual state in order for this to happen. I suspect that this issue is reproducible with sufficiently old versions of `jupyter` installed directly via `pip`, and perhaps old versions of `tqdm` and/or `ipywidgets`. However, since I couldn't reproduce this issue locally, this fix is speculative.

dimaryaz:
I can think of two solutions:
- Add the correct version of `ipywidgets` as a dependency
- Check the version at runtime, and import the plain tqdm if it's wrong

ResidentMario:
Now https://github.com/quiltdata/t4/pull/408.

PR #388, opened 2019-04-08
Have fetch mutate the package with updated data location on success

ResidentMario:
This PR addresses a specific user request: https://quiltusers.slack.com/archives/C1QMGJTT2/p1554666810033800.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -226,6 +226,12 @@ class PackageEntry(object):
         dest = fix_url(dest)
         copy_file(physical_key, dest, self.meta)
 
+        # reroot package physical keys after the copy operation succeeds
+        self.physical_keys = [dest] + self.physical_keys[1:]
+
+        return self
+
+
     def __call__(self, func=None, **kwargs):
         """
         Shorthand for self.deserialize()
@@ -452,14 +458,22 @@ class Package(object):
         """
         nice_dest = fix_url(dest).rstrip('/')
         file_list = []
-        
+        new_physical_keys = []
+
         for logical_key, entry in self.walk():
             logical_key = quote(logical_key)
             physical_key = _to_singleton(entry.physical_keys)
             new_physical_key = f'{nice_dest}/{logical_key}'
+            new_physical_keys.append(new_physical_key)
             file_list.append((physical_key, new_physical_key, entry.size, entry.meta))
 
         copy_file_list(file_list)
+
+        # reroot package physical keys after the copy operation succeeds
+        for (_, entry), new_physical_key in zip(self.walk(), new_physical_keys):
+            entry.physical_keys = [new_physical_key] + entry.physical_keys[1:]
+
+        return self
 
     def keys(self):
         """
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -229,6 +229,7 @@ class PackageEntry(object):
         # return a package reroot package physical keys after the copy operation succeeds
         # see GH#388 for context
         entry = self._clone()
+        entry.physical_keys = [dest]
         return entry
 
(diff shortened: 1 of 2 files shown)


api/python/t4/packages.py
```diff
@@ -226,6 +226,12 @@ def fetch(self, dest=None):
         dest = fix_url(dest)
         copy_file(physical_key, dest, self.meta)
 
+        # reroot package physical keys after the copy operation succeeds
+        self.physical_keys = [dest] + self.physical_keys[1:]
```
dimaryaz: I think `fetch` modifying `self` is weird. Just create and return a new entry?

api/python/t4/packages.py
```diff
@@ -452,15 +458,23 @@ def fetch(self, dest='./'):
...
+            new_physical_keys.append(new_physical_key)
             file_list.append((physical_key, new_physical_key, entry.size, entry.meta))
 
         copy_file_list(file_list)
 
+        # reroot package physical keys after the copy operation succeeds
+        for (_, entry), new_physical_key in zip(self.walk(), new_physical_keys):
+            entry.physical_keys = [new_physical_key] + entry.physical_keys[1:]
+
+        return self
```
dimaryaz: Same here. Return a new package?

ResidentMario:
I had to think about this for a while. From a user expectation perspective, it's reasonable for a user to believe that if they have one view of a package `a`, and one view of a package `b`, then the result of calling a mutator in `a` will be reflected in `b`.

However, I respect that actually implementing this behavior would cause a lot of code complexity. This is the whole "view versus copy" thing. Anyway, upon inspection `filter` already doesn't respect this behavior. I think you're right, we can defer on doing this weird thing until users state in no uncertain terms they really need it.

api/python/t4/packages.py
```diff
@@ -226,6 +226,13 @@ def fetch(self, dest=None):
         dest = fix_url(dest)
         copy_file(physical_key, dest, self.meta)
 
+        # return a package reroot package physical keys after the copy operation succeeds
+        # see GH#388 for context
+        entry = self._clone()
+        entry.physical_keys = [dest] + entry.physical_keys[1:]
```
dimaryaz: I'd just do `= [dest]`; we've never supported multiple physical keys, so no point in pretending.

api/python/t4/packages.py
```diff
@@ -452,15 +459,26 @@ def fetch(self, dest='./'):
...
             new_physical_key = f'{nice_dest}/{logical_key}'
+            new_physical_keys.append(new_physical_key)
             file_list.append((physical_key, new_physical_key, entry.size, entry.meta))
 
         copy_file_list(file_list)
 
+        # return a package reroot package physical keys after the copy operation succeeds
+        # see GH#388 for context
+        pkg = Package()
+        for (logical_key, entry), physical_key in zip(self.walk(), new_physical_keys):
```
dimaryaz: Can you just do it in the loop above, in a single pass?

(But watch out for the quoted logical key. May want to rename the variable or something.)

api/python/t4/packages.py
```diff
@@ -452,15 +459,25 @@ def fetch(self, dest='./'):
...
+            pkg.set(logical_key, physical_key)
```
dimaryaz: So I'm still worried about the performance hit here - especially since it's a regession.

Here's an idea: in the first `for` loop, do something like:
```
new_entry = entry._clone()
new_entry.physical_keys = [new_physical_key]
pkg.set(logical_key, new_entry)
```

That will just accept the new entry and skip checking the size, meta, etc.

ResidentMario: Done.

PR #393, opened 2019-04-10
Update search docs

akarve:
Update search docs, including new config.json format.
* [x] @ResidentMario can we delete the Advanced section on public search? It's dangerous and people who know AWS already know how to do it.


ResidentMario:
It's dangerous? Why?

akarve:
Dangerous because it can cause data leaks. We can address that danger in a future PR if needed.

PR #395, opened 2019-04-11
Allow user to see search mappings for a bucket

meffij:


api/python/t4/bucket.py
```diff
@@ -60,7 +57,25 @@ def config(self, config_url=None):
...
-        self._region = bucket_config.get('region')
+        # TODO: we can maybe get this from searchEndpoint or apiGatewayEndpoint
+        self._region = bucket_config.get('region', 'us-east-1')
+
+    def _get_mappings(self):
+        if not self._search_endpoint:
+            self.config()
+        return get_raw_mapping_unpacked(self._search_endpoint, self._region)
+
+    def get_user_meta_mappings(self):
```
akarve: isn't this `get_user_meta_schema`?

api/python/t4/search_util.py
```diff
@@ -112,3 +112,8 @@ def search(query, search_endpoint, limit, aws_region='us-east-1'):
                                     "bad query or a misconfigured search service.")
         setattr(exception, 'raw_response', raw_response)
         raise exception
+
+def get_raw_mapping_unpacked(endpoint, aws_region):
```
kevinemoore: Add the unpacking of the 'properties' fields here and return the cleaned up type info here instead of exposing these raw mappings.

Please also document the format of the elastic index dicts and include a link to relevant elastic docs.

PR #402, opened 2019-04-11
Fix api inconsistency with bucket.put_file

meffij:


api/python/t4/bucket.py
```diff
@@ -140,15 +140,15 @@ def put(self, key, obj, meta=None):
         all_meta.update(format_meta)
         put_bytes(data, dest, all_meta)
 
-    def put_file(self, key, path, meta=None):
+    def put_file(self, key, path, user_meta={}):
```
stevededalus: I think the signature should be consistent with `put`. Either both arguments are `user_meta` or both are `meta` and we mask the remapping from the user.

PR #403, opened 2019-04-12
Preview CSV as plain text

nl0:


nl0:
@akarve well, it seems far less digestible now, but works super fast
image

PR #405, opened 2019-04-12
Bucket.put() fixes for parquet (and other) writes

eode:
## Issues being addressed
Bucket.put() was writing invalid data to a '.parquet' file.

## Description
There are a few small changes here:
* When `bucket.put()` is called, use the extension while searching for a format to use.
* Fix method-name typo in `ParquetFormatHandler` that prevented search by type
* When a search includes extension as well as object type or metadata, sort the handlers that can handle that extension to the front of the results list
 * A small fix to `Bucket.put()` and `Bucket.put_file()` to prevent mutation of default args, and integrate some light changes on `Bucket.put()`
* Rider: fix PackageEntry.deserialize(), which had an outdated FormatRegistry search.
* Rider: A few minor docstring/comment changes

## Depends On
No dependencies.

## TODO
- [x] Unit Tests


api/python/t4/packages.py
```diff
@@ -198,13 +198,13 @@ def deserialize(self, func=None, **format_opts):
 
         pkey_ext = pathlib.PurePosixPath(unquote(urlparse(physical_key).path)).suffix
 
-        # Verify format can be handled before checking hash..
-        FormatRegistry.deserialize(data, self.meta, pkey_ext, check_only=True, **format_opts)
+        # Verify format can be handled before checking hash.  Raises if none found.
+        formats = FormatRegistry.search(None, self.meta, pkey_ext)
```
meffij: So `search` always either returns a nonempty list, or throws?

api/python/tests/test_formats.py
```diff
@@ -163,4 +163,37 @@ class Foo:
...
+
+    # FormatRegistry.search() should not *exclude* by extension, because
+    # users could make files like "data.monday" or whatever.
+    # However, it should *prioritize* by extension.
+    #
+    # From the above section, utf8_handler is formats[0].
+    formats = FormatRegistry.search(obj_type=str, ext='.json')
+    assert utf8_handler in formats
+    assert unicode_handler in formats
+    assert formats[0] is json_handler
```
stevededalus: new line at end of file.

api/python/t4/packages.py
```diff
@@ -198,13 +198,13 @@ def deserialize(self, func=None, **format_opts):
...
 
         # Verify hash before deserializing..
         self._verify_hash(data)
 
-        return FormatRegistry.deserialize(data, self.meta, pkey_ext, **format_opts)
+        return formats[0].deserialize(data, self.meta, pkey_ext, **format_opts)
```
stevededalus: is formats guaranteed to be non-empty?

api/python/t4/formats.py
```diff
@@ -134,15 +134,36 @@ def search(cls, obj_type=None, meta=None, ext=None):
             ext: The filename extension for the data
                 If given, then if other methods fail or are not specified,
                 the handler(s) for the extension `ext` will be returned.
+                If other methods do not fail, any handlers that support the
```
stevededalus: Update the comment at line 125, it seems to conflict with this line

api/python/t4/formats.py
```diff
@@ -134,15 +134,36 @@ def search(cls, obj_type=None, meta=None, ext=None):
...
             list: Formats in order of preference (latest-added first)
+
+        Raises:
+            QuiltException: if no matching formats are found
         """
-        # we want to retain order.
+        # Reasons to use lists and not sets:
+        # * we want to retain order, so recently added formats take precedence
+        # * at this scale, lists are faster than sets
         meta_fmts = cls.for_meta(meta)
```
stevededalus: nit: reorder these assignments to match the order of the args in the signature.

api/python/t4/formats.py
```diff
@@ -134,15 +134,36 @@ def search(cls, obj_type=None, meta=None, ext=None):
...
         typ_fmts = cls.for_type(obj_type)
+        ext_fmts = cls.for_ext(ext)
         fmt_name = cls._get_name_from_meta(meta)
 
+        def bump_ext_fmts(fmts):
+            """return a new list from fmts, shifting those in ext_fmts to the start"""
```
stevededalus: From a control flow/ readability point of view, can you try refactoring this and see if it improves. Something where you apply list intersection successively, i.e.:

```
ext_fmts = cls.for_ext(ext)
results = ext_fmts
if obj_type:
    ...
    results = [fmt for fmt in results if fmt in typ_fmts]
    ...
if fmt_name:
    ...
    results = [fmt for fmt in results if fmt in meta_fmts]
    ...

```

That will remove the nested `if` inside the `obj_type` check, reduce code duplication there, and removes the necessity for the `bump_ext_fmts` inner method.

api/python/t4/packages.py
eode: Yes.  This was mainly to also return a reason for the failure, if there was one.

eode:
> [...] pylint should have caught the `handles_type` typo, right?

No -- since it was a typo in the definition of an overridden method , the result was that the subclass had two methods -- `handles_typ` and `handles_type`.  The references to `ChildClass.handles_type()` would resolve to `ParentClass.handles_type()`, and are therefore technically valid.  Meanwhile `ChildClass.handles_typ()` is a method, and isn't subject to usage warnings -- so it just sits there unused.

api/python/t4/formats.py
eode: That would exclude `typ_fmts` and `meta_fmts` that don't match an extension.  We don't want to do that, because we need to be able to handle situations where:

* something looks like an extension, but is just '.something' at the end of the file -- like "available housing 2019-04-17 19:35:59.287858"
* something *is* an extension, but the user has specified a different format.

I can clean that up a bit, though, in any case.

eode: Done.

api/python/t4/formats.py
eode: done

api/python/t4/packages.py
eode: yes.  It either returns results, or raises a reason why it couldn't.

api/python/tests/test_formats.py
eode: fixed.

eode:
Did some additional pylint fixes and two minor bugfixes, but the pylint added a lot of cover and should probably be a separate PR.

eode:
Merged master into this branch to fix minor conflict (optional comma).  This change is minor, and this PR is already approved, so I'm going to go ahead and merge this back into master.

PR #409, opened 2019-04-15
Warn the user when the size of a remote object changes in between set time and push time

ResidentMario:
Currently we performance an existence check for files on S3 that we include in a `Package` via `set`. I wanted to remove this network request in the interest of transparency vis-a-vis the user expectation of what "set" does, but I backtracked on this pending further discussion b/c existence checks are highly useful (thanks @stevededalus).

However, the fact that we rely on the size of the `PackageEntry` as set at `set` time creates a footgun: we don't notice if the entry changes in size, which leads to an inaccurate progress bar.

This PR adds a warning that informs the user of changes in the size of a `PackageEntry` in between `set` time and `push` time, if such changes do in fact occur.

### DIFF
--- a/api/python/t4/data_transfer.py
+++ b/api/python/t4/data_transfer.py
@@ -634,11 +634,11 @@ def calculate_sha256(src_list, sizes, total_size):
                     old_size = humanize.naturalsize(size)
                     new_size = humanize.naturalsize(current_obj_size)
                     warnings.warn(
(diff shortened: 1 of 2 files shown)


ResidentMario:
Closing as unnecessary. Actually `set` targets a specific `versionId`, which is immune to change. My bad.

PR #410, opened 2019-04-15
Throw a warning if the file size changes in between set and push time

ResidentMario:


api/python/t4/packages.py
```diff
@@ -647,9 +647,14 @@ def _fix_sha256(self):
...
-        physical_keys = (entry.physical_keys[0] for entry in entries)
-        total_size = sum(entry.size for entry in entries)
-        results = calculate_sha256(physical_keys, total_size)
+        physical_keys = []
+        sizes = []
+        for entry in entries:
+            physical_keys.append(entry.physical_keys[0])
+            sizes.append(entry.size)
+
+        total_size = sum(sizes)
```
dimaryaz: Don't need this anymore

api/python/t4/packages.py
```diff
@@ -647,9 +647,14 @@ def _fix_sha256(self):
...
+        results = calculate_sha256(physical_keys, sizes)
```
dimaryaz: I _think_ a better API would be a list of tuples (path, size) - though I don't have a strong opinion.

But if you do leave it as is, then add an assert to `calculate_sha256` to make sure they have the same length.

api/python/t4/data_transfer.py
```diff
@@ -604,16 +605,30 @@ def get_size_and_meta(src):
...
                 with open(path, 'rb') as fd:
+                    current_file_size = pathlib.Path(path).stat().st_size
+                    if current_file_size != size:
+                        old_size = humanize.naturalsize(size)
+                        new_size = humanize.naturalsize(current_file_size)
+                        warnings.warn(
+                            f"Expected the package entry at {src!r} to be {old_size} in size, but "
+                            f"found an object which is {new_size} instead. This indicates that "
+                            f"the content of the file changed in between when you included this "
+                            f" entry in the package (via set or set_dir) and now. This should be "
```
dimaryaz: I'd just say "this is not supported"

api/python/t4/data_transfer.py
```diff
@@ -604,16 +605,30 @@ def get_size_and_meta(src):
...
     with tqdm(desc="Hashing", total=total_size, unit='B', unit_scale=True) as progress:
-        def _process_url(src):
+        def _process_url(src, size):
             src_url = urlparse(src)
             hash_obj = hashlib.sha256()
             if src_url.scheme == 'file':
                 path = pathlib.Path(parse_file_url(src_url))
+
                 with open(path, 'rb') as fd:
+                    current_file_size = pathlib.Path(path).stat().st_size
```
dimaryaz: `os.fstat(fd.fileno()).st_size`

api/python/t4/data_transfer.py
ResidentMario: It's not quite right to say `this is not supported` though, because we're throwing a warning, not an error; so clearly we *do* support this. Unless we want this to be an error?

api/python/t4/packages.py
ResidentMario: But you need `total_size` for `tqdm(desc="Hashing", total=total_size, unit='B', unit_scale=True)`?

api/python/t4/data_transfer.py
```diff
@@ -621,6 +624,19 @@ def _process_url(src):
                         hash_obj.update(chunk)
                         with lock:
                             progress.update(len(chunk))
+
+                    current_file_size = fd.tell()
+                    if current_file_size != size:
+                        old_size = humanize.naturalsize(size)
```
dimaryaz: And actually, I don't think `naturalsize` is a good idea... E.g.:
```
In [11]: humanize.naturalsize(10000)                                                                                                                                                                                            
Out[11]: '10.0 kB'

In [12]: humanize.naturalsize(10010)                                                                                                                                                                                            
Out[12]: '10.0 kB'
```

(This is a rare case, so being precise is more useful than being user-friendly)

api/python/t4/packages.py
ResidentMario: Did the latter.

api/python/t4/data_transfer.py
ResidentMario: Done.

api/python/t4/data_transfer.py
```diff
@@ -24,6 +24,7 @@
     from tqdm.autonotebook import tqdm
 
 import jsonlines
+import humanize
```
dimaryaz: Remove the import?

ResidentMario: ✔️

PR #414, opened 2019-04-16
Improvements to list_packages

ResidentMario:
This PR does three things:

1. Previously `t4.list_packages` was hard-coded to go to the `local` registry when referencing a registry with the `file` schema, which was an oversight on my part (thanks Jackson for the spot!)
2. Fixed a bug with `latest`. The current implementation naively assumes only one package has a `tophash` corresponding with `latest`, which is untrue: `build` can generate arbitrarily many sequential packages with the same `tophash` (but potentially with different physical keys). This could result in packages whose display name is `a/b:latest:latest`. This PR fixes this issue.
3. Refactored a few method name to make them easier to understand.

### DIFF
--- a/api/python/t4/api.py
+++ b/api/python/t4/api.py
@@ -266,24 +266,22 @@ def list_packages(registry=None):
                         f"{self._fmt_str(size_str, 15).rstrip(' ')}\t\n")
             return out
 
-    # get_package_registry may include '/.quilt', which Package.browse does not expect
     if registry is None or registry == 'local':
-        package_registry = None
+        registry = get_package_registry(None)
     else:
-        package_registry = fix_url(registry)
-
-    base_registry = get_package_registry(package_registry)
-    registry = base_registry.rstrip('.quilt/')
-    named_packages_path = urlparse(base_registry + '/named_packages')
-    registry_scheme = named_packages_path.scheme
+        registry = get_package_registry(fix_url(registry))
+    # the get_package_registry path includes '/.quilt', which Package.browse does not expect
+    registry = registry[:registry.rindex('.quilt')]
+    named_packages_urlparse = urlparse(registry + '.quilt/named_packages')
+    registry_scheme = named_packages_urlparse.scheme
 
     pkg_info = []
 
     if registry_scheme == 'file':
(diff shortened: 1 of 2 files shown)


api/python/t4/api.py
```diff
@@ -266,26 +266,29 @@ def create_str(self, pkg_info):
...
     if registry is None or registry == 'local':
-        base_registry = get_package_registry(None)
+        package_registry = None
     else:
-        base_registry = get_package_registry(fix_url(registry))
+        package_registry = fix_url(registry)
 
-    named_packages = base_registry + '/named_packages'
+    base_registry = get_package_registry(package_registry)
+    registry = base_registry.rstrip('.quilt/')
```
meffij: Please use a pathlib function instead. Also, strip and rstrip don't do what you think they do https://docs.python.org/2/library/string.html#string.rstrip

api/python/t4/api.py
```diff
@@ -266,26 +266,29 @@ def create_str(self, pkg_info):
...
+    named_packages_path = urlparse(base_registry + '/named_packages')
+    registry_scheme = named_packages_path.scheme
 
     pkg_info = []
 
-    registry_url = urlparse(named_packages)
-    if registry_url.scheme == 'file':
-        registry_dir = pathlib.Path(parse_file_url(registry_url))
+    if registry_scheme == 'file':
+        registry_dir = pathlib.Path(parse_file_url(named_packages_path))
```
meffij: this should be named `named_packages_dir` not `registry_dir`

ResidentMario: Done.

api/python/t4/api.py
ResidentMario: You're right, using `rstrip` is dangerous. I always think it does what I think it does and always end up re-learned that it doesn't. 🙄 However, I used `rindex` instead, not a `pathlib` function. The problem `pathlib.Path([...]).parent` is that it doesn't behave well in the presence of a path that starts with `file://` or `s3://`; it replaces it with `file:/` or `s3:/` when you convert it back using e.g. `as_posix()`, which is wrong and causes downstream operations on the path, like `parse_s3_url`, to fail. So I have to stick to string operations here.

PR #416, opened 2019-04-17
Catalog: toolchain upgrade

nl0:
Upgrade webpack to v4 and babel to v7, port changes from react-boilerplate releases 3.6, 3.7 and 4.0 (upcoming), tweak build configs and refactor some stuff. Also, remove some unused files.


nl0:
@akarve we don't need `.nginx.conf` and `.htaccess` provided by the boilerplate, do we? and what about `database.rules.json`?

PR #418, opened 2019-04-17
Exclude setup.py from codecov

ResidentMario:


ResidentMario:
@meffij can you verify that this actually does what I think it does? Because we have two code-coverage configuration files in the repo `.codecovrc` and `codecov.yml`, both of which include exclude options, so it's unclear to me whether I need to change one or both to keep codecov from failing PRs [like this one](https://github.com/quiltdata/t4/commit/ccffb3e2fb168d57368a41749da63e70dbaca242).

ResidentMario:
(per discussion, it's better to leave it out of `codecov.yml`, which controls the codecov tool, then `.codecovrc`, which controls how the `coverage` package indexes.)

PR #426, opened 2019-04-22
API enhancements for search

meffij:
Allows searching using Elastic's full Query DSL. Also allows getting the full mappings response when desired.

### DIFF
--- a/api/python/t4/search_util.py
+++ b/api/python/t4/search_util.py
@@ -76,11 +76,13 @@ def search(query, search_endpoint, limit, aws_region='us-east-1'):
 
     if isinstance(query, dict):
         payload = query
-    else:
+    elif isinstance(query, str):
         payload = {'query': {'query_string': {
             'default_field': 'content',
             'query': query,
         }}}
+    else:
+        raise QuiltException('Value provided for `query` is invalid')
 
     if limit:
         payload['size'] = limit
--- a/api/python/t4/search_util.py
+++ b/api/python/t4/search_util.py
@@ -74,10 +74,13 @@ def search(query, search_endpoint, limit, aws_region='us-east-1'):
     """
     es_client = _create_es(search_endpoint, aws_region)
 
-    payload = {'query': {'query_string': {
-        'default_field': 'content',
-        'query': query,
-    }}}
+    if isinstance(query, dict):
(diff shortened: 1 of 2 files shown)


api/python/t4/search_util.py
```diff
@@ -74,10 +74,13 @@ def search(query, search_endpoint, limit, aws_region='us-east-1'):
     """
     es_client = _create_es(search_endpoint, aws_region)
 
-    payload = {'query': {'query_string': {
-        'default_field': 'content',
-        'query': query,
-    }}}
+    if isinstance(query, dict):
+        payload = query
+    else:
```
akarve: shouldn't we check if `query` is a string at this point? if acceptable inputs are `dict` or `string` that should also land in the docstrings. we should do the most Pythonic thing. i.e. is a single argument that can assume multiple types Pythonic, or do we want to introduce, e.g. `query_dict=` as a fresh kwarg? @ResidentMario

ResidentMario:
```
if isinstance(query, dict):
    payload = query
elif isinstance(query, str):
    payload = {'query': {'query_string': {
        'default_field': 'content',
        'query': query,
    }}}
else:
    raise ValueError
```

ResidentMario: Checking types is fine. Having multiple types of input is fine, we don't actually expose the `dict`-based API to the user anyway, we only talk about the string-based API in docs so w/e. Assuming pass-through values are actually strings is unsafe and leads to poor error messages.

meffij: I took your advice to make sure the 'neither dict nor str' case raises. I agree that keeping this an undocumented feature is probably correct for now

api/python/t4/search_util.py
```diff
@@ -74,10 +74,15 @@ def search(query, search_endpoint, limit, aws_region='us-east-1'):
     """
     es_client = _create_es(search_endpoint, aws_region)
 
-    payload = {'query': {'query_string': {
-        'default_field': 'content',
-        'query': query,
-    }}}
+    if isinstance(query, dict):
+        payload = query
+    elif isinstance(query, str):
```
akarve: fine because client is not 2.7 compatible

PR #427, opened 2019-04-22
Edit setup.py to resolve dependency

meffij:


api/python/setup.py
```diff
@@ -67,6 +67,7 @@ def run(self):
         'ruamel.yaml<=0.15.70',
         'six>=1.10.0',
         'tqdm>=4.26.0',
+        'urllib3==1.24.2',             # required by requests
```
dimaryaz: I think it's a bit too strict; I'd just copy the range that `requests` specifies. (The user may already have a different but acceptable version installed, or might have another package that has weird restrictions.)

PR #428, opened 2019-04-23
Catalog: format sources w/ prettier

nl0:


nl0:
@akarve the diff is huge, but the only change is the code formatting. from now on it will be enforced automatically

PR #455, opened 2019-04-24
Renovate 1x / week

akarve:
* too many messages otherwise
* @nl0 can you be responsible for merging the weekly renovates? monthly is fine by me, too.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,7 @@
 {
   "extends": [
-    "config:base"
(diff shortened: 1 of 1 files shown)


nl0:
@akarve

> can you be responsible for merging

sure

> too many messages

only the first time, after this batch i believe it will be 1-2 per week or so,
so i'll revert this if you don't mind

PR #475, opened 2019-04-30
Harden navigator-init

meffij:


lambdas/navigator_config/index.py
```diff
@@ -97,7 +108,7 @@ def wipe_bucket(event, context):
                 bucket.delete_objects(Delete={'Objects': shard})
         cfnresponse.send(event, context, cfnresponse.SUCCESS, {})
     except Exception as e:
-        print(e)
+        print(str(e))
```
dimaryaz: It's weird that `handler` calls `wipe_bucket`, but they both look like entrypoints: they parse the event, catch all exceptions, and send responses. The check for event type being "Delete" is also redundant. Can you refactor `wipe_bucket` to make it a proper helper function?

PR #482, opened 2019-05-02
Support user paths with ~ and clean up fetch

ResidentMario:


api/python/t4/packages.py
```diff
@@ -218,9 +218,8 @@ def fetch(self, dest=None):
             None
         """
         if dest is None:
-            name = pathlib.Path(_to_singleton(self.physical_keys)).name
-            name = name.split('?versionId=')[0]
-            dest = name
+            pkey = _to_singleton(self.physical_keys)
+            dest = pathlib.PurePosixPath(unquote(urlparse(pkey).path)).name
 
         physical_key = _to_singleton(self.physical_keys)
```
dimaryaz: May want to just move this up and reuse it?

api/python/t4/packages.py
```diff
@@ -218,9 +218,8 @@ def fetch(self, dest=None):
             None
         """
         if dest is None:
-            name = pathlib.Path(_to_singleton(self.physical_keys)).name
-            name = name.split('?versionId=')[0]
-            dest = name
+            pkey = _to_singleton(self.physical_keys)
+            dest = pathlib.PurePosixPath(unquote(urlparse(pkey).path)).name
```
dimaryaz: You should then make a proper URL out of this, and move `fix_url` below into an `else:`. E.g.:
```
name = pathlib.PurePosixPath(unquote(urlparse(pkey).path)).name
dest = (pathlib.Path().resolve() / name).as_uri()
```
(If you're downloading e.g. `s3://bucket/~user`, you probably don't want to expand the `~`.)

dimaryaz: (That said, `name` could end up being `..` - that's also bad.)

ResidentMario: Name would only end up being `..` if `pathlib.PurePosixPath([...])` returns e.g. `a/b/..`, which I guess is possible but it would be _really weird_ if someone named their file `..`. 

You are right about not needing or wanting to run `fix_url` on this logic's output, I see how this would cause problems with names starting in `~` or in `..`. I moved it under an `else`.

api/python/t4/packages.py
ResidentMario: Good eye, done.

PR #483, opened 2019-05-02
Sync READMEs, indicate production ready

akarve:
* Remove ~alpha~ caveats
* Remove bad links to alpha.quiltdata.com (old stack)
* Will need to update stack links again when we have a user-facing sandbox


akarve:
> Approved for now, but we shouldn't rely on the AICS stack for this purpose for any extended period of time.

Agreed

PR #499, opened 2019-05-06
Notification configuration lambda

meffij:
Add a new lambda to set up notifications on the data bucket.

### DIFF
--- a/lambdas/notification-configuration/index.py
+++ b/lambdas/notification-configuration/index.py
@@ -11,9 +11,9 @@ def create_mappings(params):
     """ Sets up desired mappings after checking no mappings currently exist. """
     s3 = boto3.client('s3')
     existing = s3.get_bucket_notification_configuration(Bucket=params['Bucket'])
-    if existing['TopicConfigurations'] \
-            or existing['QueueConfigurations'] \
-            or existing['LambdaFunctionConfigurations']:
+    if 'TopicConfigurations' in existing \
+            or 'QueueConfigurations' in existing \
+            or 'LambdaFunctionConfigurations' in existing:
         raise Exception('Unable to attach notifications. Notification already configured '
                 'on bucket {}.'.format(params['Bucket']))
 
--- a/lambdas/notification-configuration/index.py
+++ b/lambdas/notification-configuration/index.py
@@ -21,6 +21,7 @@ def create_mappings(params):
     return
 
 def handler(event, context):
+    """ Top-level handler for custom resource """
     print('Changing bucket notification settings')
     try:
         params = dict(event['ResourceProperties'])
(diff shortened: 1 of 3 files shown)


lambdas/notification-configuration/index.py
```diff
@@ -0,0 +1,66 @@
...
+            return
+        else:
+            # unknown event type
+            send(event, context, FAILED, physical_resource_id=current_resource_id,
+                    reason='Unknown event type ' + event['RequestType'])
+            return
+
+    except Exception as e:
+        print('Exception encountered')
+        print(str(e))
```
stevededalus: Can we print `event` here too just for the sake of having as much debug info as possible?

lambdas/notification-configuration/index.py
```diff
@@ -0,0 +1,66 @@
...
+
+    s3.put_bucket_notification_configuration(**params)
+    return
+
+def handler(event, context):
+    """ Top-level handler for custom resource """
+    print('Changing bucket notification settings')
+    try:
+        params = dict(event['ResourceProperties'])
+        del params['ServiceToken']
```
stevededalus: Is it better to black list the `ServiceToken` field or whitelist all other fields? Can you add a comment to that effect? Seems safer to explicitly copy fields.

lambdas/notification-configuration/index.py
```diff
@@ -0,0 +1,66 @@
...
+        elif event['RequestType'] == 'Update':
+            if event['PhysicalResourceId'] == current_resource_id:
+                # do nothing if physical_resource_id is unchanged
+                send(event, context, SUCCESS, physical_resource_id=current_resource_id)
+                return
+            # Otherwise, bucket name changed. Must set up new notification.
+            # Also must delete old notification on old bucket.
+            create_mappings(params)
+            # delete old notification
+            params = dict(event['OldResourceProperties'])
```
stevededalus: let's call this `old_params` just to be a little clearer on the two cases

lambdas/notification-configuration/index.py
```diff
@@ -0,0 +1,66 @@
...
+            del params['ServiceToken']
+            params['NotificationConfiguration'] = {}
+            s3.put_bucket_notification_configuration(**params)
+            # report success with new resource id
+            send(event, context, SUCCESS, physical_resource_id=current_resource_id)
+            return
+        elif event['RequestType'] == 'Delete':
+            # do nothing
+            send(event, context, SUCCESS, physical_resource_id=current_resource_id)
```
stevededalus: should we not be deleting the notifications here the same way you do in update?

lambdas/notification-configuration/index.py
```diff
@@ -0,0 +1,66 @@
+"""
+Set up indexer notification on data bucket.
+
+Remove notification on delete.
+"""
+import boto3
+
+from t4_lambda_shared.cfnresponse import send, SUCCESS, FAILED
+
+def create_mappings(params):
```
stevededalus: let's call this `update_mappings` `put_mappings` or `add_mappings` so its clear that its making a remote call and not just setting a request up.

PR #518, opened 2019-05-13
Stop catching the "same file" exception

dimaryaz:
Not sure what the original reason for catching it was, but:
- It's completely avoidable: we can check if paths are equal and we're not setting metadata
- The logic is broken because we can't return a versioned URL
- We're already raising an exception when copying a local file onto itself


dimaryaz:
Ping

PR #524, opened 2019-05-15
Handle s3:TestEvent (and other unexpected events)

akarve:
* Make indexer more robust to unexpected event types
* Slight readability tweaks

### DIFF
--- a/lambdas/es/indexer/index.py
+++ b/lambdas/es/indexer/index.py
@@ -168,8 +168,17 @@ def handler(event, _):
     dispatch post_to_es
     """
     try:
-        for msg in event['Records']:
-            for record in json.loads(json.loads(msg['body'])['Message'])['Records']:
+        # existence of event['Records'] is guaranteed by
+        # https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html
+        # see above for details on event structure, s3:TestEvent
+        for outer in event['Records']:
+            # these should all be SNS messages (indexer consumes from SNS topic, by design)
+            raw_message = json.loads(outer['body'])['Message']
+            message = json.loads(raw_message)
+            if 'Records' not in message:
+                # consume event (we don't want to index it)
+                return
+            for record in message['Records']:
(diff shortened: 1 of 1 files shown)


lambdas/es/indexer/index.py
```diff
@@ -168,8 +168,17 @@ def handler(event, _):
...
-            for record in json.loads(json.loads(msg['body'])['Message'])['Records']:
+        # existence of event['Records'] is guaranteed by
+        # https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html
+        # see above for details on event structure, s3:TestEvent
+        for outer in event['Records']:
+            # these should all be SNS messages (indexer consumes from SNS topic, by design)
+            raw_message = json.loads(outer['body'])['Message']
+            message = json.loads(raw_message)
+            if 'Records' not in message:
+                # consume event (we don't want to index it)
```
kevinemoore: Are we sure to be the only consumer of the events in this Queue? If not, it might be worth writing a special case for TestEvent and simply ignoring (not consuming) other events.

akarve: @stevededalus is there a way to ignore events without consuming them? from what i read below, consumption == lambda task did not throw.

lambdas/es/indexer/index.py
```diff
@@ -185,7 +194,8 @@ def handler(event, _):
                     elif eventname == 'ObjectCreated:Put':
                         event_type = 'Create'
                     else:
-                        event_type = eventname
+                        # we should only send either Create or Delete events
+                        return
```
meffij: Event handling is kind of weird in the indexer before your change, but we need to handle many more event types than what are explicitly mentioned. https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html#notification-how-to-event-types-and-destinations

meffij: Returning without handling these events will lead to objects not being indexed correctly

PR #533, opened 2019-05-15
Fix commit messages

dimaryaz:
Not only is the code broken - but it comes with a unit test that pactches the broken function and passes!

Stop patching things.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -950,7 +950,6 @@ class Package(object):
             A new package that points to the copied objects.
         """
         validate_package_name(name)
-        self._set_commit_message(message)
 
         if registry is None:
             if dest is None:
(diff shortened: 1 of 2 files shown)


dimaryaz:
Also, @akarve: the catalog code appears to be broken, too - it doesn't get the commit message from the package. (But of course you can't really test it until this one is fixed.)

api/python/t4/packages.py
```diff
@@ -950,7 +950,6 @@ def push(self, name, dest=None, registry=None, message=None):
             A new package that points to the copied objects.
         """
         validate_package_name(name)
-        self._set_commit_message(message)
```
stevededalus: Should we delete the message definition too? looks like it isn't called anywhere else

PR #537, opened 2019-05-16
Multi-bucket support for notification configuration lambda

meffij:
Just no-ops on an empty string bucket. Will be useful for adding multi-bucket support to the template (current plan is to create one copy of each custom resource for each bucket, but we'll need a placeholder value in case users have less than the maximum number of allowed buckets. Empty string is a good sentinel value).


stevededalus:
Is the design documented somewhere? This seems pretty undeveloped: https://docs.google.com/document/d/1ohAb136FGtGAyVbqKu6rBE9_f1DalRsMN52paQH89IQ/edit

PR #546, opened 2019-05-20
Add a lambda that calculates object and package statistics

dimaryaz:
It uses Athena queries to get object access counts and a mapping of package hashes to names/versions,
then manually joins them and outputs three summaries: `OBjects.csv`, `Packages.csv`, `PackageVersions.csv`.

It would be nice to use Parquet - but pyarrow is too big for lambdas. In the future, we could try
using temporary tables in Athena to create Parquet output.

### DIFF
--- a/lambdas/access_counts/index.py
+++ b/lambdas/access_counts/index.py
@@ -144,9 +144,9 @@ def handler(event, context):
         s3.delete_object(Bucket=QUERY_RESULT_BUCKET, Key=key + '.metadata')
 
     outputs = [
-        ('Objects.csv', object_access_counts, ['eventname', 'bucket', 'key']),
-        ('Packages.csv', package_access_counts, ['eventname', 'bucket', 'name']),
-        ('PackageVersions.csv', package_version_access_counts, ['eventname', 'bucket', 'name', 'version', 'hash']),
+        ('Objects.csv', object_access_counts, ['eventname', 'bucket', 'key', 'counts']),
+        ('Packages.csv', package_access_counts, ['eventname', 'bucket', 'name', 'counts']),
+        ('PackageVersions.csv', package_version_access_counts, ['eventname', 'bucket', 'name', 'version', 'hash', 'counts']),
     ]
 
     for name, data, header in outputs:
--- a/lambdas/access_counts/index.py
+++ b/lambdas/access_counts/index.py
@@ -1,0 +1,161 @@
+"""
+Lambda function that runs Athena queries over CloudTrail logs and .quilt/named_packages/
+and creates summaries of object and package access events.
+"""
+
+import csv
+from collections import defaultdict
+import json
+import os
+import time
+from tempfile import TemporaryFile
+
+import boto3
+
+# A saved query that reduces CloudTrail logs event counts per object per day.
+OBJECT_ACCESS_COUNTS_QUERY_ID = os.environ['OBJECT_ACCESS_COUNTS_QUERY_ID']
+# Named packages
+NAMED_PACKAGES_QUERY_ID = os.environ['NAMED_PACKAGES_QUERY_ID']
+# Bucket where query results will be stored.
+QUERY_RESULT_BUCKET = os.environ['QUERY_RESULT_BUCKET']
+# A temporary directory where Athena query results will be written.
+QUERY_TEMP_DIR = os.environ['QUERY_TEMP_DIR']
+# Directory where the summary files will be stored.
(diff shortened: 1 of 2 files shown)


lambdas/access_counts/index.py
```diff
@@ -0,0 +1,161 @@
...
+NAMED_PACKAGES_QUERY_ID = os.environ['NAMED_PACKAGES_QUERY_ID']
+# Bucket where query results will be stored.
+QUERY_RESULT_BUCKET = os.environ['QUERY_RESULT_BUCKET']
+# A temporary directory where Athena query results will be written.
+QUERY_TEMP_DIR = os.environ['QUERY_TEMP_DIR']
+# Directory where the summary files will be stored.
+ACCESS_COUNTS_OUTPUT_DIR = os.environ['ACCESS_COUNTS_OUTPUT_DIR']
+
+
+MAX_RESULTS = 1000
```
kevinemoore: Make an environment variable so we can change without redeploying?

lambdas/access_counts/index.py
```diff
@@ -0,0 +1,161 @@
...
+    # (eventname, bucket, package_name) -> {date: count, ...}
+    package_access_counts = defaultdict(lambda: defaultdict(int))
+
+    # (eventname, bucket, package_name, package_version, package_hash) -> {date: count, ...}
+    package_version_access_counts = defaultdict(lambda: defaultdict(int))
+
+    named_packages_execution_id = run_query(NAMED_PACKAGES_QUERY_ID)
+    access_counts_execution_id = run_query(OBJECT_ACCESS_COUNTS_QUERY_ID)
+
+    named_packages_key = '%s/%s.csv' % (QUERY_TEMP_DIR, named_packages_execution_id)
```
akarve: no need to change this, but i don't understand why we wouldn't use f-strings for `.format`.

lambdas/access_counts/index.py
```diff
@@ -0,0 +1,161 @@
...
+    access_counts_key = '%s/%s.csv' % (QUERY_TEMP_DIR, access_counts_execution_id)
+
+    wait_for_query(named_packages_execution_id)
+
+    for bucket, name, version, hash in get_query_results(named_packages_execution_id):
```
akarve: `hash` is shadowing  `hash`; maybe `hash_`?

lambdas/access_counts/index.py
```diff
@@ -0,0 +1,161 @@
...
+    header = True
+
+    while True:
+        response = athena.get_query_results(**args)
+        for row in response['ResultSet']['Rows']:
+            if header:
+                # Athena returns the header before the actual results. Sigh.
+                header = False
+            else:
+                yield [value['VarCharValue'] for value in row['Data']]
```
akarve: it would be super helpful to document (in this file) the expected schemas of all underlying files. for instance, i was looking at https://github.com/quiltdata/t4/pull/546/files#diff-76a245204f5971faf14b2f56e0b84da4R130 and i have no clear sense of whether that line actually matches the data we're returning.

lambdas/access_counts/index.py
```diff
@@ -0,0 +1,161 @@
...
+    ]
+
+    for name, data, header in outputs:
+        with TemporaryFile('w+') as file_obj:
+            write_counts_csv(file_obj, data, header)
+            file_obj.seek(0)
+            s3.put_object(
+                Body=file_obj.buffer.raw,
+                Bucket=QUERY_RESULT_BUCKET,
+                Key=ACCESS_COUNTS_OUTPUT_DIR + '/' + name,
```
akarve: not `os.path.join`? i realize this is an s3 path

lambdas/access_counts/index.py
```diff
@@ -0,0 +1,161 @@
...
+        next_token = response.get('NextToken')
+        if next_token is None:
+            break
+        else:
+            args.update(
+                NextToken=next_token
+            )
+
+
+def write_counts_csv(fd, data, field_names):
```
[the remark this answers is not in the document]
dimaryaz: Uhh, no, it's writing CSV. One of the fields is JSON.

lambdas/access_counts/index.py
dimaryaz: Exactly. `os.path.join` is for OS paths.

lambdas/access_counts/index.py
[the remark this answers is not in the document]
akarve: my bad; i see it

PR #547, opened 2019-05-21
Add a lambda that calculates object and package statistics

dimaryaz:
It uses Athena queries to get object access counts and a mapping of package hashes to names/versions, joins them, and outputs three summaries: `Objects.csv`, `Packages.csv`, `PackageVersions.csv`.

It would be nice to use Parquet - but pyarrow is too big for lambdas.

### DIFF
--- a/lambdas/access_counts/index.py
+++ b/lambdas/access_counts/index.py
@@ -57,7 +57,7 @@ CREATE_PACKAGE_HASHES = textwrap.dedent(f"""\
     SELECT
         -- Parse a file path like `s3://BUCKET/.quilt/named_packages/USER_NAME/PACKAGE_NAME/VERSION`
         split_part("$path", '/', 3) AS bucket,
-        split_part("$path", '/', 6) || '/' || split_part("$path", '/', 7) AS name,
+        concat(split_part("$path", '/', 6), '/', split_part("$path", '/', 7)) AS name,
         split_part("$path", '/', 8) AS version,
         hash
     FROM named_packages
@@ -93,7 +93,7 @@ PACKAGE_VERSION_ACCESS_COUNTS = textwrap.dedent("""\
         hash,
         CAST(histogram(date) AS JSON) AS counts
     FROM object_access_log JOIN package_hashes
-    ON object_access_log.bucket = package_hashes.bucket AND key = '.quilt/packages/' || hash
+    ON object_access_log.bucket = package_hashes.bucket AND key = concat('.quilt/packages/', hash)
     GROUP BY eventname, package_hashes.bucket, name, version, hash
 """)
 
--- a/lambdas/access_counts/index.py
+++ b/lambdas/access_counts/index.py
@@ -1,0 +1,200 @@
+"""
+Lambda function that runs Athena queries over CloudTrail logs and .quilt/named_packages/
+and creates summaries of object and package access events.
+"""
+
+import os
+import textwrap
+import time
+
+import boto3
+
+ATHENA_DATABASE = os.environ['ATHENA_DATABASE']
+# Bucket where query results will be stored.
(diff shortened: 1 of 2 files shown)


lambdas/access_counts/index.py
```diff
@@ -0,0 +1,200 @@
...
+    SELECT eventname, date_format(from_iso8601_timestamp(eventtime), '%Y-%m-%d') AS date, bucket, key
+    FROM (
+        SELECT
+            eventname,
+            eventtime,
+            json_extract_scalar(requestparameters, '$.bucketName') AS bucket,
+            json_extract_scalar(requestparameters, '$.key') AS key
+        FROM cloudtrail
+    )
+    WHERE bucket IS NOT NULL AND key IS NOT NULL
```
akarve: is this WHERE clause insurance or are there cases when bucket or key can be null? if the latter, worth a comment.

lambdas/access_counts/index.py
```diff
@@ -0,0 +1,200 @@
...
+CREATE_PACKAGE_HASHES = textwrap.dedent(f"""\
+    CREATE TABLE package_hashes
+    WITH (
+        format = 'Parquet',
+        parquet_compression = 'SNAPPY',
+        external_location = 's3://{sql_escape(QUERY_RESULT_BUCKET)}/{sql_escape(QUERY_TEMP_DIR)}/package_hashes/'
+    )
+    AS
+    SELECT
+        split_part("$path", '/', 3) AS bucket,
```
akarve: comment with expected path, e.g. `s3://BUCKET/AWSLogs/730278974607/CloudTrail/us-east-1/2019/04/01/730278974607_CloudTrail_us-east-1_20190401T0220Z_ZLJUquHVvB845TeS.json.gz` will make these splits easier to read and maintain.

lambdas/access_counts/index.py
```diff
@@ -0,0 +1,200 @@
...
+                ContinuationToken=list_response['ContinuationToken']
+            ))
+        else:
+            break
+
+
+def handler(event, context):
+    delete_temp_dir()
+
+    for query_id in [run_query(DROP_OBJECT_ACCESS_LOG), run_query(DROP_PACKAGE_HASHES)]:
```
akarve: worth noting that this is safe since it happens in an athena db owned by the stack? (i.e. `object_access_logs` could be a common table name).

lambdas/access_counts/index.py
```diff
@@ -0,0 +1,200 @@
...
+""")
+
+PACKAGE_ACCESS_COUNTS = textwrap.dedent("""\
+    SELECT
+        eventname,
+        package_hashes.bucket AS bucket,
+        name,
+        CAST(histogram(date) AS JSON) AS counts
+    FROM object_access_log JOIN package_hashes
+    ON object_access_log.bucket = package_hashes.bucket AND key = '.quilt/packages/' || hash
```
akarve: i don't follow `key = '.quilt/packages/' || hash`; can you clarify?

lambdas/access_counts/index.py
dimaryaz: Well, CloudTrail can contain all kinds of events - user logged in, etc. - not even S3 related. (But even S3 events like ListBucket don't have a key.)

lambdas/access_counts/index.py
dimaryaz: Ok

lambdas/access_counts/index.py
dimaryaz: `||` just concatenates strings

lambdas/access_counts/index.py
dimaryaz: Ok

lambdas/access_counts/index.py
akarve: ah, i'd find `concat()` more readable

PR #549, opened 2019-05-21
Implement full parallelization of uploads and remote copies.

dimaryaz:


api/python/t4/data_transfer.py
```diff
@@ -126,44 +126,61 @@ def _upload_file(callback, size, src_path, dest_bucket, dest_key, override_meta)
...
-        with open(src_path, 'rb') as fd:
-            for i in itertools.count(1):
-                # TODO(dima): Better progress callback.
-                chunk = fd.read(s3_transfer_config.multipart_chunksize)
-                if not chunk:
-                    break
-                part = s3_client.upload_part(
-                    Body=chunk,
+
+        chunk_offsets = list(range(0, size, s3_transfer_config.multipart_chunksize))
```
stevededalus: Can you write a test where the chucksize is very small to verify it produces enough and correct upload parts?

PR #552, opened 2019-05-22
Disallow "." and ".." file names and folder names in package keys

ResidentMario:
Objects with the name `.` or `..` create serious issues in our client because even though these are legal S3 object names, these names have special meaning on the local filesystem. It's not sanitary to use these as file names or folder names even inside of S3.

We currently allow creating such objects and working with such objects, but we need to stop doing so because they're a foot gun.

This PR prevents users from creating packages containing `.` or `..` paths. A follow-up PR will do the same for the bucket API.

### DIFF
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -852,7 +852,6 @@ class Package(object):
 
         if isinstance(entry, (string_types, getattr(os, 'PathLike', str))):
             url = fix_url(str(entry))
-            validate_key(url)
             size, orig_meta, version = get_size_and_meta(url)
 
             # Deterimine if a new version needs to be appended.
@@ -863,7 +862,6 @@ class Package(object):
                     url = make_s3_url(bucket, key, version)
             entry = PackageEntry([url], size, None, orig_meta)
         elif isinstance(entry, PackageEntry):
-            validate_key(_to_singleton(entry.physical_keys))
             entry = entry._clone()
         else:
             raise TypeError(
--- a/api/python/t4/packages.py
+++ b/api/python/t4/packages.py
@@ -23,7 +23,7 @@ from .formats import FormatRegistry
 from .util import (
     QuiltException, fix_url, get_from_config, get_install_location,
     get_package_registry, make_s3_url, parse_file_url, parse_s3_url,
-    validate_package_name, quiltignore_filter
+    validate_package_name, quiltignore_filter, validate_key
 )
 
 
@@ -838,7 +838,12 @@ class Package(object):
             self
         """
         if not logical_key or logical_key.endswith('/'):
-            raise QuiltException("Invalid logical_key: %r; cannot be a directory" % logical_key)
+            raise QuiltException(
+                f"Invalid logical key {logical_key!r}. "
+                f"A package entry logical key cannot be a directory."
+            )
+
+        validate_key(logical_key)
 
         if not entry:
             current_working_dir = pathlib.Path.cwd()
@@ -846,7 +851,9 @@ class Package(object):
             entry = logical_key_abs_path.relative_to(current_working_dir)
 
         if isinstance(entry, (string_types, getattr(os, 'PathLike', str))):
-            url = fix_url(str(entry))
+            entry_str = str(entry)
+            validate_key(entry_str)
+            url = fix_url(entry_str)
             size, orig_meta, version = get_size_and_meta(url)
 
             # Deterimine if a new version needs to be appended.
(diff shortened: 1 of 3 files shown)


api/python/t4/packages.py
```diff
@@ -838,15 +838,22 @@ def set(self, logical_key, entry=None, meta=None):
...
+        validate_key(logical_key)
 
         if not entry:
             current_working_dir = pathlib.Path.cwd()
             logical_key_abs_path = pathlib.Path(logical_key).absolute()
             entry = logical_key_abs_path.relative_to(current_working_dir)
 
         if isinstance(entry, (string_types, getattr(os, 'PathLike', str))):
-            url = fix_url(str(entry))
+            entry_str = str(entry)
```
dimaryaz: These are physical keys - shouldn't validate those.

It's totally reasonable to call `pkg.set('foo', '../../foo.txt')`, and `fix_url` will take care of that by resolving relative paths. It won't do anything about proper URLs that contain `/../`... but that's a different problem. Technically, they're legal URLs, and shouldn't break anything. We can test that.

api/python/t4/packages.py
```diff
@@ -857,9 +864,13 @@ def set(self, logical_key, entry=None, meta=None):
                     url = make_s3_url(bucket, key, version)
             entry = PackageEntry([url], size, None, orig_meta)
         elif isinstance(entry, PackageEntry):
+            validate_key(_to_singleton(entry.physical_keys))
```
dimaryaz: Same here.

api/python/t4/util.py
```diff
@@ -366,3 +366,19 @@ def quiltignore_filter(paths, ignore, url_scheme):
...
+def validate_key(key):
+    """
+    Verify that a file path or S3 path does not contain any '.' or '..' separators or files.
+    """
+    if key is None or key == '':
+        raise QuiltException(
+            f"Invalid key {key!r}. A package entry key cannot be empty."
+        )
+
+    if ((key == '.' or '/./' in key or key.startswith('./') or key.endswith('/.')) or
```
dimaryaz: That's kind of ugly... Do something like this instead: https://github.com/quiltdata/t4/blob/master/api/python/t4/data_transfer.py#L513

Or even just move that function into `util.py`

api/python/t4/util.py
```diff
@@ -366,3 +366,19 @@ def quiltignore_filter(paths, ignore, url_scheme):
         return files.union(dirs)
     else:
         raise NotImplementedError
+
+def validate_key(key):
+    """
+    Verify that a file path or S3 path does not contain any '.' or '..' separators or files.
+    """
+    if key is None or key == '':
```
dimaryaz: `if not key`

ResidentMario:
I don't like the idea of allowing physical keys that point to files or objects with bad names like `.` and `..` because in a follow-up PR I want to prevent users from writing objects to these physical keys in the bucket API, and it would be inconsistent if these physical keys were allowed in the packages API but not the bucket API:

```python
import t4

# this would be allowed
t4.Package().set("LOGICAL_KEY", "s3://foo/bar/.")

# this would be disallowed
t4.Bucket("s3://foo").put_file("s3://foo/bar/.", "LOCAL_FILEPATH")
```

If we have any restrictions around what kinds of S3 object names are legal in our API, I would like for those restrictions to be consistent across both the `Package` and `Bucket` API.

Or, put another way, I don't want packages to have "allowed badness" as a feature.

So for this reason I think the restriction on the naming of the physical key is appropriate. This implementation doesn't quite get it right though because I neglected to consider the local path expansion that `fix_url` does! I will fix this.

api/python/t4/packages.py
ResidentMario: The only URLs that we currently accept are of the following form:

```
file://$USERPATH/$FILENAME
s3://$BUCKET/$OBJECT_PATH
```

If `$USERPATH`, `$FILENAME`, or `$OBJECT_PATH` contains a `.` or `..` path fragment that's an error. If `$BUCKET` is `.` or `..`, then you have a bucket named `.` or `..`, which seems impossible?

api/python/t4/packages.py
ResidentMario: Note that since we regard setting an entry as a `PackageEntry` as a special case, I thought about leaving this check out, but I couldn't because `set_dir` (surprisingly, in my opinion) uses this code path.

ResidentMario:
After further discussion we decided to drop the physical key validation. The argument is that it's better to mess less with what the user is trying to do than to enforce semantic consistency, which I buy.

api/python/t4/util.py
ResidentMario: Done.

api/python/t4/packages.py
ResidentMario: Resolved per in-person discussion.

ResidentMario:
> Also need to check:
> * logical key in `set_dir`
> * logical keys in existing packages (e.g., if you access someone else's package that contains invalid keys)

Re 1: `set_dir` ultimately calls `root.set(logical_key, entry)`, so `validate_key` does get called there.

Re 2: packages in the wild don't have this problem because we know such packages don't exist, so I don't think defensive code is necessary here? I also don't know that we can do anything in this case.

